PR #12, opened 2024-07-04
Changed Experiment to use interpolation as well

aecker:


experanto/data.py
```diff
@@ -5,22 +5,21 @@
...
-        chunk_idx = slice(s, s + self.chunk_size)
-        return list(self._experiment[chunk_idx].values())
-
-    def get_device_names(self):
-        return self._experiment.device_names
-    
\ No newline at end of file
+        times = self._sample_times[s:s+self.chunk_size]
+        data, _ = self._experiment.interpolate(times)
+        return list(data.values())
```
fabiansinz, line 25: does that ensure that we always get the values in the correct order? I think it might be best to transition to `namedtuple` for return values.

experanto/experiment.py
```diff
@@ -5,74 +5,73 @@
...
+class Experiment:
 
-class Experiment(Sequence):
-
-    def __init__(self, root_folder: str, sampling_rate: float) -> None:
+    def __init__(self, root_folder: str) -> None:
         self.root_folder = Path(root_folder)
-        self.sampling_rate = sampling_rate
         self._devices = dict()
         self.device_names = list()
```
fabiansinz, line 14: That seems redundant and a possible cause for errors. I would suggest to make it a property:

```python
@property
def device_names(self):
    return tuple(self._devices.keys())
```

experanto/experiment.py
```diff
@@ -5,74 +5,73 @@
...
         self._load_devices()
 
-
     def _load_devices(self) -> None:
         # Populate devices by going through subfolders
         # Assumption: blocks are sorted by start time
         device_folders = [d for d in self.root_folder.iterdir() if d.is_dir()]
         self.device_names = [f.name for f in device_folders]
         for d in device_folders:
             print("Parsing {} data... ".format(d.name), end="")
```
fabiansinz, line 25: Printing should go into logging soon.

fabiansinz (approved):

PR #14, opened 2024-07-05
images time dim added

pollytur:


aecker:
@pollytur can you resolve the conflicts, please? 

Also, it would be good to clear the output of the Jupyter notebook before committing.

Issue #18, opened 2024-07-05
add staticImageDataset


KonstantinWilleke:

aecker:
A first version is now implemented. However, it probably won't be plug & play yet. The following will need to be fixed (at least) for it to work with legacy neuralpredictors models:

- Add the color channel at the right place for the `screen` device
- Add behavior as input channels at the right place
- Potentially rename devices (in constructor?) to be consistent with legacy models

@KonstantinWilleke @pollytur can one of you either point me to an example model or make these changes?

aecker:
@pollytur I am sorry but I didn't get to do the changes w.r.t. making it compatible with legacy models before my vacation. Could you take over?

KonstantinWilleke:
I'm also happy to help

pollytur:
As the screen dataset already handles static images and behaviour - I close this issue

PR #23, opened 2024-07-05
Implemented static image dataset

aecker:
Implements issue #18


experanto/data.py
```diff
@@ -30,3 +32,49 @@ def __getitem__(self, idx):
...
+        screen = self._experiment.devices["screen"]
+        self._trials = [
+            t for t in screen.trials
+            if isinstance(t, ImageTrial) and t.get_meta("tier") == self.tier
+        ]
+        s_idx = np.array([t.first_frame_idx for t in self._trials])
+        if len(s_idx):
+            self._start_times = screen.timestamps[s_idx]
+            end_times = screen.timestamps[s_idx + 1]
+            self._trial_duration = np.min(end_times - self._start_times)
```
KonstantinWilleke, line 57: @aecker I feel a bit uneasy about setting the trial duration to `np.min(end_times - self._start_times)`.

The minimum time could maybe be a few shorter than average. If you use that as the averaging window for all images, that could lead to problems.

I like though that it's data driven and does not rely on the meta info. For the images, we have the "stim duration=0.5" stored in each image meta file. 

I had a similar problem with adding the very last blank. Because the very last flip time does not exist, I just added the most common difference in time between two frames. But that's probably overkill. 
Looks like that:
```
digits = 4
most_common_time_diff_between_frames = np.unique(np.round(np.diff(all_flip_times), digits),
                    return_counts=True)[0][0]
```
the unique is returning as element [0][0] the most common occurence.

experanto/data.py
```diff
@@ -30,3 +32,49 @@ def __getitem__(self, idx):
...
+            assert np.all(
+                end_times - self._start_times - self._trial_duration < 0.1), \
+                "Trial durations vary by more than 100 ms"
```
KonstantinWilleke, line 60: we should make this assumption a bit more explicit somewhere. I think there might be some experiments where the stimulus duration was more unequal.

Now that I think about it - maybe this has to be a user specified argument. Offset + duration of time window could be user specified. If we add an assert as explicit data sanity check, then we need too many Dataset classes for all kinds of experiments.

KonstantinWilleke (changes requested):

experanto/data.py
KonstantinWilleke: see comment below for my change of heart

aecker:
@KonstantinWilleke I made `stim_duration` an input and removed the check for equal stimulus durations. After reconsidering it, I realized that the assert will probably cause more headaches than solve issues. It would make more sense to do the sanity checking at export time.

KonstantinWilleke (approved):

KonstantinWilleke:
> @KonstantinWilleke I made `stim_duration` an input and removed the check for equal stimulus durations. After reconsidering it, I realized that the assert will probably cause more headaches than solve issues. It would make more sense to do the sanity checking at export time.

yes fully agreed!

Issue #25, opened 2024-07-07
add dynamic dataset 


KonstantinWilleke:
for comparing model performance on senorium2023 data

pollytur:
this also requires implementing neuron motor coordinates and unit id as properties (experiment.device.coord and experiment.device.unit_ids)

pollytur:
for the track of records - we agreed on linear interpolation for sequence interpolator and to add the concept of valid times and flag if we want to return nans or non

pollytur:
again, as ChunckDataset handles videos and images - this is not valid anymore

Issue #32, opened 2024-07-16
namedtuple is incompatible with num_workers>0 in PyTorch DataLoader


KonstantinWilleke:
The `__getitem__` of the datasets can not handle a namedtuple when there are multiple parallel workers. The parallel workers are needed to reach high data loading speeds for powerful compute nodes.
This is a fundamental pytorch issue: because each worker instantiates a `Dataset` object, and the namedtuple is instantiated in each one, the parallel workers can't collate the batches in a "custom" named tuple. 

Potential workarounds: 
- write a custom `collate_fn` for the DataLoader which converts the entire batch into a namedtuple. 
- each batch returns a dictionary, as in `{device_name: value_tensor}`

pollytur:
Another potential solution is to do [same as in neuralpredictors](https://github.com/sinzlab/neuralpredictors/blob/5e86a8222db48b6eab271e8ff672fa67b29012f8/neuralpredictors/data/datasets/base.py#L75) and define the namedtuple globally, then it works with the num_workers>0

pollytur:
this is still valid
named tuples needs to be removed from 
* https://github.com/sensorium-competition/experanto/blob/ee767fb34b82ee0a8569e4a2099efbf0b5d396c6/experanto/experiment.py#L5 
* https://github.com/sensorium-competition/experanto/blob/ee767fb34b82ee0a8569e4a2099efbf0b5d396c6/experanto/datasets.py#L7 
* `self.DataPoint` also needs to be removed - https://github.com/sensorium-competition/experanto/blob/ee767fb34b82ee0a8569e4a2099efbf0b5d396c6/experanto/datasets.py#L55
* I am not sure about `TimeInterval` - https://github.com/sensorium-competition/experanto/blob/ee767fb34b82ee0a8569e4a2099efbf0b5d396c6/experanto/intervals.py#L7 - we do not use it so far in the data loading but might become an issue later

PR #39, opened 2024-09-07
Video dataset

pollytur:
ok, linear interpolation is now added 
but as for the valid times - I would actually not go this way as there are a lot of open questions on how to define validity now (is it 3 or 5 points which should be not missing in a row for valid interpolation, is it 1 behavioural channel out of 2 or is it all of them, etc)
So probably indeed makes sense to let this open for the modeling side to decide, as Fabian suggested
Same reason not to replace things with 0 and adding validity flag - because its not clear if we want to do it per channel or overall, do we want to consider things across modalities or not, etc

related to #25 but not fully as this video dataset primarily done to match the sensorium dataset 
It would not allow to load chunck with 2 videos in 1 chance yet (because it requires much more complex valid time logic and this need discussion is we want to go down this path)

### DIFF
--- a/experanto/data.py
+++ b/experanto/data.py
@@ -93,6 +93,7 @@ class Mouse2pVideoDataset(Dataset):
         cut: bool,
         add_channel: bool,
         channel_pos: int,
+        rescale: bool = False,
     ) -> None:
         """
         this dataloader returns the full the video resampled to the new freq rate
@@ -111,12 +112,13 @@ class Mouse2pVideoDataset(Dataset):
         self.tier = tier
         self.sampling_rate = sampling_rate
         self.stim_duration = stim_duration
-        self._experiment = Experiment(root_folder)
+        self._experiment = Experiment(root_folder, rescale=rescale)
         self.device_names = self._experiment.device_names
         self.subsample = subsample
         self.cut = cut
         self.add_channel = add_channel
         self.channel_pos = channel_pos
+        self.rescale = rescale
         assert (
             0 <= channel_pos < 4
         ), "channels could be extended only for positions [0,3]"
@@ -160,7 +162,7 @@ class Mouse2pVideoDataset(Dataset):
         :return: this would return video in data['screen'] with shape of [t, h, w]
         """
         fs = self.sampling_rate
-        times = np.arange(self._start_times[idx], self._end_times[idx], fs)
+        times = np.arange(self._start_times[idx], self._end_times[idx], 1/fs)
         # get all times possible
         # cut is needed
         if self.cut:
--- a/experanto/data.py
+++ b/experanto/data.py
@@ -114,6 +114,10 @@ class Mouse2pVideoDataset(Dataset):
         self.stim_duration = stim_duration
         self._experiment = Experiment(root_folder, rescale=rescale)
         self.device_names = self._experiment.device_names
+        # this is needed to match sensorium order only
+        if 'screen' in self.device_names and 'responses' in self.device_names:
+            start = ['screen', 'responses']
+            self.device_names = tuple(start) + tuple(set(self.device_names).difference(set(start)))
         self.subsample = subsample
         self.cut = cut
         self.add_channel = add_channel
@@ -176,5 +180,8 @@ class Mouse2pVideoDataset(Dataset):
 
         if self.add_channel and len(data["screen"].shape) != 4:
             data["screen"] = np.expand_dims(data["screen"], axis=self.channel_pos)
+        # this hack matches the shape for sensorium models  
+        if 'responses' in data:
+            data['responses'] = data['responses'].T
 
-        return data
+        return self.DataPoint(**data)
--- a/experanto/data.py
+++ b/experanto/data.py
@@ -12,21 +12,34 @@ from .experiment import Experiment
 from .interpolators import ImageTrial, VideoTrial
 
 DEFAULT_INTERP_CONFIG = {
-    "screen": {"rescale": False},
+    "screen": {
+        "rescale": True,
+        "normalize": True,
+    },
     "responses": {
         "keep_nans": False,
+        #                            'interpolation_mode' : 'nearest_neighbor',
         "interpolation_mode": "linear",
         "interp_window": 5,
+        "normalize": True,
+        "normalize_subtract_mean": False,
+        "normalize_std_threshold": 0.01,
     },
     "eye_tracker": {
         "keep_nans": False,
         "interpolation_mode": "linear",
         "interp_window": 5,
+        "normalize": True,
+        "normalize_subtract_mean": True,
+        "normalize_std_threshold": None,
     },
     "treadmill": {
         "keep_nans": False,
         "interpolation_mode": "linear",
         "interp_window": 5,
+        "normalize": True,
+        "normalize_subtract_mean": True,
+        "normalize_std_threshold": None,
     },
 }
 
@@ -249,4 +262,7 @@ class Mouse2pVideoDataset(Dataset):
 
         if self.add_channel and len(data["screen"].shape) != 4:
             data["screen"] = np.expand_dims(data["screen"], axis=self.channel_pos)
+        # this hack matches the shape for sensorium models
+        if "responses" in data:
+            data["responses"] = data["responses"].T
         return self.DataPoint(**data)
--- a/experanto/data.py
+++ b/experanto/data.py
@@ -11,6 +11,25 @@ from torch.utils.data import Dataset
 from .experiment import Experiment
 from .interpolators import ImageTrial, VideoTrial
 
+DEFAULT_INTERP_CONFIG = {
+    "screen": {"rescale": False},
+    "responses": {
+        "keep_nans": False,
+        "interpolation_mode": "linear",
+        "interp_window": 5,
+    },
+    "eye_tracker": {
+        "keep_nans": False,
+        "interpolation_mode": "linear",
+        "interp_window": 5,
+    },
+    "treadmill": {
+        "keep_nans": False,
+        "interpolation_mode": "linear",
+        "interp_window": 5,
+    },
+}
+
 
 class Mouse2pChunkedDataset(Dataset):
     def __init__(
@@ -18,20 +37,14 @@ class Mouse2pChunkedDataset(Dataset):
         root_folder: str,
         sampling_rate: float,
         chunk_size: int,
-        rescale: bool = False,
-        keep_nans: bool = False,
-        interpolation_mode: str = "linear",
-        interp_window: int = 5,
+        interp_config: dict = DEFAULT_INTERP_CONFIG,
     ) -> None:
         self.root_folder = Path(root_folder)
         self.sampling_rate = sampling_rate
         self.chunk_size = chunk_size
         self._experiment = Experiment(
             root_folder,
-            rescale=rescale,
-            keep_nans=keep_nans,
-            interpolation_mode=interpolation_mode,
-            interp_window=interp_window,
+            interp_config,
         )
         self.device_names = self._experiment.device_names
         self.start_time, self.end_time = self._experiment.get_valid_range("screen")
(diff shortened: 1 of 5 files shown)


experanto/data.py
```diff
@@ -3,20 +3,36 @@
...
+        chunk_size: int,
+        rescale: bool = False,
+        keep_nans: bool = False,
+        interpolation_mode: str = "linear",
+        interp_window: int = 5,
+    ) -> None:
         self.root_folder = Path(root_folder)
         self.sampling_rate = sampling_rate
         self.chunk_size = chunk_size
-        self._experiment = Experiment(root_folder)
```
KonstantinWilleke, line 19: I think here the cleanest way is to have a config dict for all Interpolators.
The config dict would be an input kwarg to the Dataset, and then passed to Experiment and down to the interpolators.

e.g.:
`interp_config = {"screen": {...}, {"eye_tracker": {...}, }`

and then it'd look like this: 
`self._experiment = Experiment(root_folder, interp_config)`

Otherwise the args/kwargs for Experiment() are out of control when more devices and functionality are added

experanto/experiment.py
```diff
@@ -27,15 +40,23 @@ def _load_devices(self) -> None:
 
         for d in device_folders:
             log.info(f"Parsing {d.name} data... ")
-            dev = Interpolator.create(d)
```
KonstantinWilleke, line 30: With regard to my config dict suggestion above,  it will simplify to:


    def _load_devices(self) -> None:
        # Populate devices by going through subfolders
        # Assumption: blocks are sorted by start time
        device_folders = [d for d in self.root_folder.iterdir() if d.is_dir()]

        for d in device_folders:
            if d not in self.interp_config:
                continue
            log.info(f"Parsing {d.name} data... ")
            dev = Interpolator.create(d, self.interp_config)
            self.devices[d.name] = dev
            self.start_time = dev.start_time
            self.end_time = dev.end_time
            log.info("Parsing finished")

experanto/interpolators.py
```diff
@@ -82,6 +101,17 @@ def __init__(self, root_folder: str) -> None:
                 self.start_time + np.max(self._phase_shifts),
                 self.end_time + np.min(self._phase_shifts),
             )
+        # todo - make it lazy loading? and read-only properties?
+        if "neuron_properties" in meta:
```
KonstantinWilleke, line 105: I'd rather put all of this into the `Dataset` class, to keep the interpolators simple and clean. I'd rather give the dataset all the fancy properties, and keep the functionality of the interpolators to the basic.

it can look like this in the dataset:
f = self._experiment.devices["responses"].root_folder + "/meta/unit_ids.npy"
if os.exist(f):
   self.unit_ids = np.load(f)

experanto/data.py
```diff
@@ -3,20 +3,36 @@
...
 
 
 class Mouse2pChunkedDataset(Dataset):
-    def __init__(self, root_folder: str, sampling_rate: float, chunk_size: int) -> None:
+    def __init__(
+        self,
+        root_folder: str,
+        sampling_rate: float,
+        chunk_size: int,
+        rescale: bool = False,
```
KonstantinWilleke, line 21: The rescaling could/should be something that the `transforms` of the dataset get_item handle, or is there a case where it needs to be done at the level of loading the single frame?

experanto/data.py
```diff
@@ -3,20 +3,36 @@
...
+        keep_nans: bool = False,
```
KonstantinWilleke, line 22: the other kwargs can be dealt with the config_dict that I describe below see below

experanto/data.py
```diff
@@ -36,25 +52,36 @@ def __getitem__(self, idx):
...
-            tier: str, 
-            offset: float, 
-            stim_duration: float
-            ) -> None:
+        self,
+        root_folder: str,
+        tier: str,
+        offset: float,
+        stim_duration: float,
+        rescale: bool = False,
```
KonstantinWilleke, line 60: same as above

experanto/data.py
```diff
@@ -36,25 +52,36 @@ def __getitem__(self, idx):
...
+        interp_window: int = 5,
+    ) -> None:
         self.root_folder = Path(root_folder)
         self.tier = tier
         self.offset = offset
         self.stim_duration = stim_duration
-        self._experiment = Experiment(root_folder)
+        self._experiment = Experiment(
+            root_folder,
+            rescale=rescale,
```
KonstantinWilleke, line 71: same as above

experanto/data.py
```diff
@@ -36,25 +52,36 @@ def __getitem__(self, idx):
...
+            keep_nans=keep_nans,
+            interpolation_mode=interpolation_mode,
+            interp_window=interp_window,
+        )
         self.device_names = self._experiment.device_names
         self.DataPoint = namedtuple("DataPoint", self.device_names)
```
KonstantinWilleke, line 77: As related to #32, I'd suggest to not doing any datapoints and rather stick with dicts

experanto/data.py
```diff
@@ -74,8 +101,141 @@ def __getitem__(self, idx):
...
+        """
+
+        :param idx: idx of the video
+        :return: this would return video in data['screen'] with shape of [t, h, w]
+        """
+        fs = self.sampling_rate
+        times = np.arange(self._start_times[idx], self._end_times[idx], 1 / fs)
+        # get all times possible
+        # cut is needed
+        if self.cut:
```
KonstantinWilleke, line 226: I dont fully understand this here, but maybe we can chat about this

experanto/experiment.py
```diff
@@ -1,23 +1,36 @@
...
-import logging
 
 log = logging.getLogger(__name__)
 
 
 class Experiment:
+    def __init__(
+        self,
+        root_folder: str,
+        rescale: bool = False,
```
KonstantinWilleke, line 20: same as above regarding the config dict

experanto/interpolators.py
```diff
@@ -52,22 +55,38 @@ def __contains__(self, times: np.ndarray):
...
+    def __init__(
+        self,
+        root_folder: str,
+        keep_nans: bool = False,
+        # interpolation_mode: str = 'nearest_neighbor',
+        interpolation_mode: str = "linear",
+        interp_window: int = 5,
+    ) -> None:
+        """
+        interpolation_mode - nearest meighbor or linear
```
KonstantinWilleke, line 80: typo:)

experanto/interpolators.py
```diff
@@ -113,17 +143,85 @@ def interpolate(self, times: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
...
+                )
+                assert (
+                    array.shape[0] == orig_times.shape[0]
+                ), "times and data should be same length before interpolation"
+                data = linear_interpolate_sequences(
+                    array, orig_times, valid_times, self.keep_nans
+                )
 
-        return data, valid
+            else:
```
KonstantinWilleke, line 176: not sure I get it, what other case than nearest neighbor or linear is there?

KonstantinWilleke (changes requested):

experanto/data.py
pollytur: I agree conceptually but we need to discuss it in a bigger crowd I guess because I am not sure if we agreed on how to make the transforms logic (so far I put it here as the datasets I exported are with different dimensions)

experanto/interpolators.py
pollytur: this `else` is still under `elif self.interpolation_mode == "linear":`
and its
```
if len(idx.shape) == 1:
    ...
else:
```
because the `start_idx` and `end_idx` needs to be taken along dimensions now

experanto/data.py
pollytur: this is mostly for sensorium backcompatibility tests 
there we subsample like N frames from each video during training
but during validation we take the whole video
so if cut is false - it would return the whole video, like during validation

experanto/interpolators.py
```diff
@@ -113,17 +143,85 @@ def interpolate(self, times: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
...
+                    )
+                    assert (
+                        local_data.shape[0] == local_time.shape[0]
+                    ), "times and data should be same length before interpolation"
+                    data[n_idx] = linear_interpolate_1d_sequence(
+                        local_data, local_time, valid_times, self.keep_nans
+                    )
+            #                 data = data.T
+
+            return data, valid
```
pollytur, line 207: @pollytur -todo remove frames added for interpolations around

pollytur:
@KonstantinWilleke thanks a lot for the review! I will fix the configulations and move loading of neural properties to the dataset and then we can discuss in person what is left:)

experanto/data.py
KonstantinWilleke: Ok I understand! I see that it makes sense to have a dataset that reproduces sensorium exactly.

In the general case, for a general purpose video dataset, this could be a also added to the list of transforms. but let's do that after we come up with a good principle

experanto/interpolators.py
KonstantinWilleke: Oh, I'm stupid. yes makes sense

KonstantinWilleke:
Hi, I've cloned your branch and tried it out! It seems that there are some performance issues that I can't identify how they originate.
With the new default interp config, the dataloader is very slow, 25x slower than the previous version, both with num_workers=0. 

With a config changed so that it should reproduce the old version, it's still 50% slower.
This is the config:
```
config = {
    "screen": {
        "rescale": False,
    },
    "responses": {
        "keep_nans": False,
        #                            'interpolation_mode' : 'nearest_neighbor',
        "interpolation_mode": "nearest_neighbor",
    },
    "eye_tracker": {
        "keep_nans": False,
        "interpolation_mode": "nearest_neighbor",
    },
    "treadmill": {
        "keep_nans": False,
        "interpolation_mode": "nearest_neighbor",
    },
}
```
could you also try it out on your end?

pollytur:
@KonstantinWilleke thanks for checking! Will test and come back to you.

pollytur:
I tested and profiled some thing, so the main slow down is linear interpolation
Does it match with your numbers @KonstantinWilleke ?

I used `stim_duration=30, sampling_rate=30, batch_size = 32, num_workers=0`

```
Mouse2pVideoDataset, linear interpolation
1682364 function calls (1681963 primitive calls) in 2.646 seconds

Mouse2pVideoDataset, knn interpolation
13496 function calls (13095 primitive calls) in 0.602 seconds

Mouse2pChunkedDataset, linear interpolation
1682337 function calls (1681873 primitive calls) in 2.749 seconds

Mouse2pChunkedDataset, knn interpolation
13146 function calls (12823 primitive calls) in 0.836 seconds
```

pollytur:
`scipy.interpolate.interp1d` is supposed to be deprecated soon, thats why I did not used it, though I am trying it now for the speed up
I will test the previous version as well
And yes, its the for loop that is the problem, I am still working on it.

KonstantinWilleke:
I've tested the speed again and I have to take back my earlier statement. For the config that reproduces the "old" dataloader, the speed is identical.

I was running this:
```
config = {
    "screen": {
        "rescale": True,
        "normalize": False,
    },
    "responses": {
        "keep_nans": True,
        "interpolation_mode": "linear",
        "normalize": False,
    },
    "eye_tracker": {
        "keep_nans": True,
        "interpolation_mode": "linear",
        "normalize": False,
    },
    "treadmill": {
        "keep_nans": True,
        "interpolation_mode": "linear",
        "normalize": False,
    },
}
```

on:
```
root_folder = "/data/mouse_datasets/dynamic29228-2-10-Video-sensorium23_full_256_144/"
sampling_rate = 30  # Hz
chunk_size = 120 # samples per block, i.e. context length in samples per neuron
dataset = Mouse2pChunkedDataset(root_folder=root_folder, chunk_size=chunk_size, sampling_rate=sampling_rate, interp_config=config)
```
that was just as fast as the old branch. ~2mins per epoch for num_workers =0. 
With linear interpolation it is 6min per epoch.

pollytur:
ok, then its ~3 times difference for both of us. Thanks for testing.
I would try to speed up the for loop, and ping you again.

KonstantinWilleke:
ok, thanks a lot, yes that makes sense! I'll try now if the multi-animal dataloader is feasible with separate workers per animal.

pollytur:
as we moved to the Chunck dataset and it does covers the functionality from this PR is a more efficient ways, I close this PR

PR #40, opened 2024-10-14
Multi-session and finished ChunkDataset

KonstantinWilleke:
This PR has a new ChunkDataset. It can return chunks based on arbitrary conditions (tiers, stim_types, ...).
It also introduces the following features:
- a default .yaml file for configurations based on hydra
- simple multisession loader construction
- all transforms bases on torchvision.transforms run in get_item
- variable sampling rates, chunk_sizes, and offsets for each modality
- best practices for fast dataloader speed (achieves >= 10,000frames per second of loading)

It includes a breaking change to the VideoDataloader @pollytur, so maybe you can just test out the chunkloader. Then we can discuss of how to make them both useable.


pollytur:
Nice, thanks a lot. I will test today and let you know

experanto/dataloaders.py
```diff
@@ -0,0 +1,17 @@
...
+from torch.utils.data import DataLoader
+
+from .datasets import ChunkDataset
+from .utils import MultiEpochsDataLoader, LongCycler
+
+
+def get_multisession_dataloader(paths, config: DictConfig) -> DataLoader:
+    dataloaders = {}
+    for i, path in enumerate(paths):
+        dataset_name = path.split("dynamic")[1].split("-Video")[0]
```
pollytur, line 12: this would limit the name folders, why just not to take the whole folder name? Then it would not have to contain "dynamic", etc in it and it might be easier to integrate Allen brain dataset, for instance

experanto/interpolators.py
```diff
@@ -113,17 +168,95 @@ def interpolate(self, times: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
...
+                    local_time = (
+                        np.arange(st_idx, end_idx[n_idx]) * self.time_delta
+                        + self.valid_interval.start
+                    )
+                    assert (
+                        local_data.shape[0] == local_time.shape[0]
+                    ), "times and data should be same length before interpolation"
+                    out = linear_interpolate_1d_sequence(
+                        local_data, local_time, valid_times, self.keep_nans
+                    )
```
pollytur, line 229: this should be deleted. It actually my error, sorry for that

experanto/datasets.py
```diff
@@ -0,0 +1,448 @@
...
+                        ToTensor(),
+                        torchvision.transforms.Normalize(self._statistics[device_name]["mean"], self._statistics[device_name]["std"])])
+                else:
+                    transforms[device_name] = Compose([
+                        ToTensor()])
+        return transforms
+
+    def get_sample_in_meta_condition(self) -> dict:
+        """
+        iterates through all
```
pollytur, line 388: it took me a bit of time to get through it, so I would suggest to add a bit more info in the doc strings, see the suggestion below
```suggestion
        iterates through all stimuli, selects the ones which match the meta conditions (tiers or stimuli types) and creates a mask to select the correct times using `self._screen_sample_times` as the clock/reference times
```

experanto/datasets.py
```diff
@@ -0,0 +1,448 @@
...
+                if condition == v or (condition == "blank" and self.modality_config["screen"]["include_blanks"]):
+                    sample_in_meta.append(
+                        (self._screen_sample_times >= start) & (self._screen_sample_times < end)
+                    )
+            sample_in_meta_condition[k] = np.stack(sample_in_meta).sum(0).astype(bool)
+        return sample_in_meta_condition
+
+    def get_full_valid_sample_times(self, ) -> Iterable:
+        """
+        iterates through all sample times and checks if they are in the meta condition of interest
```
pollytur, line 407: it took me a bit of time to get through it, so I would suggest to add a bit more info in the doc strings, see the suggestion below
```suggestion
        iterates through all sample times and checks if they could be used as the start times, eg if the next `self.chunk_sizes["screen"]` points are still valid based on the previous meta condition filtering
```

experanto/datasets.py
```diff
@@ -0,0 +1,448 @@
...
+        self._read_trials()
+        self.initialize_statistics()
+        self._screen_sample_times = np.arange(
+            self.start_time, self.end_time, 1.0 / self.sampling_rates["screen"]
+        )
+        # iterate over the valid condition in modality_config["screen"]["valid_condition"] to get the indices of self._screen_sample_times that meet all criteria
+        self._sample_in_meta_condition = self.get_sample_in_meta_condition()
+        self._full_valid_sample_times = self.get_full_valid_sample_times()
+        # the _valid_screen_times are the indices from which the starting points for the chunks will be taken
+        self._valid_screen_times = self._full_valid_sample_times[::self.sample_stride]
```
pollytur, line 337: This might be a bit dangerous in general scenario, as if the chunk and the sample_stride are big we might throw away too much data. 
I do not have a good solution suggestion, so a warning could help at least to make users aware.

experanto/datasets.py
```diff
@@ -0,0 +1,448 @@
...
+            self.meta_conditions[k] = [t.get_meta(k) if t.get_meta(k) is not None else "blank" for t in self._trials]
+
+    def initialize_statistics(self) -> None:
+        self._statistics = {}
+        for device_name in self.device_names:
+            self._statistics[device_name] = {}
+            # If modality should be normalized, load respective statistics from file.
+            if self.modality_config[device_name].transforms.get("normalization", False):
+                mode = self.modality_config[device_name].transforms.normalization
+                assert mode in ['standardize', 'normalize']
```
pollytur, line 357:
```suggestion
                assert mode in ['standardize', 'normalize'], f"transforms.normalization should be in 'standardize' or 'normalize', not {mode}"
```

experanto/datasets.py
```diff
@@ -0,0 +1,448 @@
...
+            maybe_all_true.append(correct_duration)
+            for k, v in self._sample_in_meta_condition.items():
+                maybe_all_true.append(np.all(v[i: i + self.chunk_sizes["screen"]]))
+            if np.all(maybe_all_true):
+                valid_times.append(self._screen_sample_times[i])
+        return np.stack(valid_times)
+
+    def shuffle_valid_screen_times(self) -> None:
+        """
+        convenience function to randomly select new starting points for each chunk. Use this in training after each epoch.
```
pollytur, line 423: maybe its a nice idea to add more comments (from slack with Mara) on why this is here

experanto/interpolators.py
```diff
@@ -40,7 +43,7 @@ def __init__(self, root_folder: str) -> None:
 
     def load_meta(self):
         with open(self.root_folder / "meta.yml") as f:
-            meta = yaml.load(f, Loader=yaml.FullLoader)
+            meta = yaml.load(f, Loader=yaml.Loader)
```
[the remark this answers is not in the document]
KonstantinWilleke: Oh I see, yes will add that

experanto/datasets.py
KonstantinWilleke: Yes, I'm very sorry that I commented so sparsely. Will commit yours and add more.

experanto/datasets.py
KonstantinWilleke: Yes agreed. The behavior of this should be made clearer. 

I'd argue that for our case, mostly these two cases are relevant:
- stride=1, this is for training, so that every sample can be a valid starting point
- stride = chunk_size, without randomization. This would mean that every frame gets sampled once, like in the validation and test set

experanto/datasets.py
KonstantinWilleke: Yes agreed. I thought that this is a good idea to add this function, but for all practical purposes, the training should be `stride=1` anyhow got get as much data as possible. so now I'm thinking that this function isn't super useful, but might have its use cases

pollytur:
@KonstantinWilleke let me know once I am supposed to look at it again :) 
(I am not sure based on the comments)

experanto/datasets.py
KonstantinWilleke: I added more docstrings now

experanto/datasets.py
KonstantinWilleke: I added more comments here.

KonstantinWilleke:
@pollytur Could you review again? I've made all the changes :)

pollytur:
LGTM now, thanks a lot!

One question before the merge - Yongrong mentioned some bug with the `sample_stride != 1`. Is it somehow addressed here or this is not an issue anymore

PR #58, opened 2025-02-13
timestamps_fix

pollytur:
There was a small bug before, if we use different frequences/chunk sizes before then the times for the last modality in the for loop would be used as times for the responses are not necessarily the times from responses modality

I also checked with Yongrong and Alex - it should not affect the current response-to-response model as they used same sampling rate in the dataloader

I now return all times for multimodal models


pollytur:
@KonstantinWilleke can you look through it? Or is there anyone on your side who alternatively can look through it?
I dont want to merge it silently as it would break dependencies with Alex and Yongrong perceiver repo ('timestamps' are now 'responses_timestamps')

pollytur:
the filter branch PR has this functionality fixed as well now, so I close this PR

PR #61, opened 2025-04-08
Mouse foundation model

KonstantinWilleke:
This PR makes necessary changes in order to make dataloaders using experanto more robust, lightweight, and efficient.

Features added:
- utility for combining all meta.yaml into a single json for faster reading speed
- utility for data caching
- making sure that all tensor are returned as .contiguous() tensors 
- making sure that all tensors are instantiated and manipulated as float32
- add filters for NaNs
- add filter for artefacts 
- full redesign how datasets are combined in a pytorch dataloader, via using torch.ConcatDataset. This removes Cyclers around dataloaders which is more efficient and allows for easier State tracking. This solves all performance related issues in a large scale distributed context, is more performant, and avoids worker contention in torch.multiprocessing


experanto/dataloaders.py
```diff
@@ -1,17 +1,131 @@
...
 
 
-def get_multisession_dataloader(paths, config: DictConfig) -> DataLoader:
     dataloaders = {}
-    for i, path in enumerate(paths):
-        dataset_name = path.split("dynamic")[1].split("-Video")[0]
-        dataset = ChunkDataset(path, **config.dataset,)
+    for i, (path, cfg) in enumerate(zip(paths, configs)):
+        # TODO use saved meta dict to find data key
+        if "dynamic" in path:
```
fabiansinz, line 41: Do we really want to tie configurations to file path names. That looks like a hack that will eventually break is an super hard to convey in the documentation.

experanto/dataloaders.py
```diff
@@ -1,17 +1,131 @@
...
+
+    # Create datasets
+    datasets = []
+    session_names = []
+
+    start_time = time.time()
+    for i, (path, cfg) in enumerate(zip(paths, configs)):
+        # Extract session name
+        if "dynamic" in path:
+            session_name = path.split("dynamic")[1].split("-Video")[0]
```
fabiansinz, line 90: Same here + duplicate code.

experanto/datasets.py
```diff
@@ -394,59 +436,189 @@ def initialize_transforms(self):
...
+                        print(visualization_string)
+                    if valid_intervals is None:
+                        valid_intervals = valid_intervals_
+                    else:
+                        valid_intervals = find_intersection_between_two_interval_arrays(valid_intervals, valid_intervals_)
+
+        return valid_intervals
+    
+        
+    def get_condition_mask_from_meta_conditions(self, valid_conditions_sum_of_product: List[dict]) -> np.ndarray:
```
fabiansinz, line 500: I feel we should also compare how they do it in the temporal data of the Dyer-Lab.

fabiansinz: That's not required right now but a comment for later.

experanto/dataloaders.py
KonstantinWilleke: Agreed. I left both `dataloaders` convenience functions in, mostly for comparison and benchmarking.

(1) `get_multisession_dataloader` is using the LongCyler logic.
(2) `get_multisession_concat_dataloader` is using the torch.ConcatDataset logic.

Happy to get rid of (1) entirely, and with that all the cycler/Longcycler/Exhauster business.

experanto/dataloaders.py
KonstantinWilleke: Yes 100% agreed. Here I get a `data_key` from the path. It's not used for any configuration, but just to get a `str` for the data_key.
We should settle for a 100% robust solution.
My proposal is that we put a .json or .yaml in the top level directory of the exported dataset. This should contain the full config of the exporter. And a data key, or information of how to build a key.

In the foundation model export, here's what I did.

```
base_directory = dynamic21608-2-18-Video-b448915027fa4aa145acc0c00199ff15_30h
/screen
/responses
/eye_tracker
/...
meta.json

...
> cat meta.json

{
    "config": {
        "include_eye": true,
        "include_treadmill": true,
        "resize_stimuli": true,
        "screen_h_w": [
            36,
            64
        ],
        "tracking_method": 2,
        "mask_type": "soma",
        "trace_source": "Activity",
        "stack_coordinates": false,
        "include_astrocytes": false,
        "astrocyte_channel": 1,
        "astrocyte_video_h_w": null,
        "neuro_trace_segmentation_method": 6,
        "neuro_trace_spike_method": 5,
        "neuro_channel": 1,
        "include_anatomy": false,
        "sample_dtype": "float64",
        "screen_dtype": "uint8",
        "interleave_value": 128,
        "n_trials_to_export": null,
        "train_size": 0.8,
        "val_size": 0.1,
        "test_size": 0.1,
        "add_blank_at_the_end": true,
        "pupil_lowpass_fr": 2
    },
    "scan_key": {
        "animal_id": "21608",
        "session": 2,
        "scan_idx": 18
    }
}
```

KonstantinWilleke: What pattern would you suggest for getting the `data_key` of a session?
Previously (in neuralpredictors) the datasets did this manually.

Now, our datasets are much more general. But it shouldn't be the dataset's job to come up with a data_key. I feel that there's no elegant solution, we just have to pick one that sticks.

KonstantinWilleke: I added a quick commit of how it could look like:
[add convenience function in datasets to infer data_key from root_dire…](https://github.com/sensorium-competition/experanto/pull/61/commits/2b2062cf35a32ed8a70647e17a4797dc25cb3ec3)

README.md
pollytur, line 1: not for now but for refactoring afterwards
If we want people outside of our direct labs network to use the package then data structure should be in the readme with some minimal example (eg as it was in the old readme)

configs/default.yaml
```diff
@@ -21,40 +25,51 @@ dataset:
...
       offset: 0
       transforms:
         normalization: "normalize"
       interpolation:
         interpolation_mode: "nearest_neighbor"
+      filters:
+        nan_filter:
+          __target__: experanto.filters.common_filters.nan_filter
+          __partial__: True
+          vicinity: 0.05
```
pollytur, line 67: I would suggest to rename it in sth line `vicinity_in_sec` to make it clear that vicinity is measured in seconds and its the only place we use wall time in the config

docker-compose.yml
```diff
@@ -1,12 +1,11 @@
...
     volumes:
-      - ./notebooks:/home/jovyan/
-      - ~/data:/home/jovyan/data/
-      - .:/src/experanto
\ No newline at end of file
+      - .:/notebooks
+    environment:
+      - JUPYTER_ENABLE_LAB=yes
+    runtime: nvidia
+    shm_size: 256g
```
pollytur, line 11: maybe also for future discussion - do we really need 256g here as a default for shared memory

setup.py
pollytur, line 1: for future refactoring, maybe move from `setup.py` to `pyproject.toml` (both can also work together)

scripts/run_stateful_dataloader.ipynb
pollytur, line 1: I am not sure if this notebook should stay or go. But if it stays it should be in `notebooks/tests/` and not in `scripts/`

scripts/dataset_filter.ipynb
pollytur, line 1: I think this notebook should be removed as a) it does not really show much b) uses `tensorfabrik` and a custom hardcoded path, so no-one actually could run it

scripts/filter_playground.ipynb
pollytur, line 1: I am not quite sure what the notebook is supposed to show, should probably go away as well

experanto/dataloaders.py
```diff
@@ -1,17 +1,128 @@
...
+        paths (List[str]): List of paths to the datasets.
+        configs (Union[DictConfig, Dict, List[Union[DictConfig, Dict]]]): Configuration for each dataset.
+            If a single config is provided, it will be applied to all datasets.
+            If a list is provided, it should match the length of paths.
+        shuffle_keys (bool): Whether to shuffle the keys of the dataloaders.
+        **kwargs: Additional keyword arguments for dataset and dataloader configuration.
+    """
+
+    if configs is None and "config" in kwargs:
+        configs = kwargs.pop("config")
```
pollytur, line 34:
```suggestion 
    if configs is None:
        if "config" in kwargs:
            configs = kwargs.pop("config")
        else:
            raise ValueError("Either 'configs' parameter must be provided or 'config' must be defined in kwargs")
    elif "config" in kwargs:
        raise ValueError("Cannot specify both 'configs' parameter and 'config' in kwargs")   
```

experanto/dataloaders.py
```diff
@@ -1,17 +1,128 @@
...
+
+    # Convert single config to list for uniform handling
+    if isinstance(configs, (DictConfig, dict)):
+        configs = [configs] * len(paths)
```
pollytur, line 38:
```suggestion
    if isinstance(configs, (DictConfig, dict)):
        configs = [configs] * len(paths)
    assert len(configs) == len(paths), 'Configs should be same length as path if passed as list'
```

scripts/run_stateful_dataloaders.py
pollytur, line 1: this should be in `tests/` and not `scripts/` and should be clearly refactored later (eg line 79 or 138 `print("\nStatefulLongCycler test passed!")`)

scripts/combine_metadatas.py
```diff
@@ -0,0 +1,165 @@
...
+from pathlib import Path
+import re
+import yaml
+import typing
+
+# Ensure correct working directory if running from a different location
+# script_dir = Path(__file__).parent.absolute()
+# project_root = script_dir.parents[1] # Adjusted from [2] to [1]
+# os.chdir(project_root)
+os.chdir("/src/code/Enigma/Libraries/PRS/tensorfabrik") # User override
```
pollytur, line 15: hardcoded path for a non-public repo, probably should be removed

scripts/combine_metadatas.py
```diff
@@ -0,0 +1,165 @@
...
+    except Exception as e:
+        print(f"An error occurred while processing {screen_folder}: {e}")
+        # Re-raise the exception to be caught by the calling script
+        raise
+# --- End of Metadata Combining Function ---
+
+
+# Define the base directory containing session folders
+# --- IMPORTANT: SET YOUR BASE DIRECTORY HERE ---
+BASE_DATA_DIR = "/mnt/stor02/enigma/full_foundation_export"
```
pollytur, line 88: same here, hardcoded path for a particular server

scripts/combine_metadatas.py
pollytur, line 1: Is this file used anywhere? I did not find it in the src. If not, I would suggest to remove it as it has hardcoded pathes, does not have clear entry point (eg `if __name__ == "__main__":`) and has values defined all over the place (between print and functions).

Unless its used somewhere or needed for some other reason - lets remove it

scripts/check_sessions.ipynb
pollytur, line 1: I would remove this ipynb as it does not show much more compared to the other ipynbs which are already there

pollytur, line 1: I would remove this ipynb as it does not show much more compared to the other ipynbs which are already there

scripts/check_sessions.py
pollytur, line 1: I think this is a nice script to check folders, let's move it to `tests/` right now and refactor later as it needs to be more in a function style with taking the path as an input rather than hardcoding it

experanto/intervals.py
```diff
@@ -0,0 +1,208 @@
+import typing
+from typing import List
+import numpy as np
+class TimeInterval(typing.NamedTuple):
```
pollytur, line 4:
```suggestion
import numpy as np

class TimeInterval(typing.NamedTuple):
```

experanto/dataloaders.py
```diff
@@ -1,17 +1,128 @@
...
+    # Create datasets
+    datasets = []
+    session_names = []
+
+    start_time = time.time()
+    for i, (path, cfg) in enumerate(zip(paths, configs)):
+
+        # Create dataset with deterministic seed
+        path_hash = hash(path) % 10000
+        dataset_seed = seed + path_hash if seed is not None else None
```
pollytur, line 94: why to have the hash part in seeds?
I think this is not great as the hashes for pathes are different lets say if data is saved in different folders or even is symlinks are used. And this creates unnecessary confusion as the seed that is used is not actually the one, which is specified
```suggestion
        dataset_seed = seed if seed is not None else None
```

experanto/datasets.py
```diff
@@ -254,7 +265,12 @@ def __init__(
         root_folder: str,
         global_sampling_rate: None,
         global_chunk_size: None,
+        add_behavior_as_channels: bool = False,
+        replace_nans_with_means: bool = False,
+        cache_data: bool = False,
+        out_keys: Optional[Iterable] = None,
```
pollytur, line 271: for future - maybe add `in_keys` as well, as we might want to load only video and responses but not behaviour and that might save quite a bit of cpu

experanto/datasets.py
```diff
@@ -362,16 +388,36 @@ def initialize_statistics(self) -> None:
...
                     means = np.zeros_like(means)
+                elif mode == 'recompute_responses':
+                     means = np.zeros_like(means)
+                     stds = np.nanstd(self._experiment.devices["responses"]._data, 0)[None, ...]
+                elif mode == 'recompute_behavior':
+                     means = np.nanmean(self._experiment.devices[device_name]._data, 0)[None, ...]
+                     stds = np.nanstd(self._experiment.devices[device_name]._data, 0)[None, ...]
+                elif mode == 'screen_default':
+                     means = np.array((80))
+                     stds = np.array((60))
```
pollytur, line 409:
```suggestion
                elif mode == 'screen_default':
                     means = np.array((80))
                     stds = np.array((60))
                else:
                    raise NotImplementedError(f'Mode {mode} is not implemented, please choose from standardize/recompute_responses/screen_default or implement yourself')
```

scripts/run_distributed_dataloaders.py
```diff
@@ -0,0 +1,412 @@
...
+        local_items.extend([("__pad__", -1) for _ in range(max_items - len(local_items))])
+
+    # For each item, gather its key and neuron count
+    all_keys = []
+    all_neurons = []
+
+    for i in range(max_items):
+        if i < len(local_items):
+            key, neurons = local_items[i]
+            key_tensor = torch.tensor([ord(c) for c in key] + [0] * (50 - len(key)), device="cuda")
```
pollytur, line 145: what is `50` here and in line 148. Could we me make it a variable with some type of a meaningful name?

experanto/datasets.py
```diff
@@ -394,59 +440,238 @@ def initialize_transforms(self):
...
+        possible_indices = np.arange(n_samples)
+        
+        # Check duration condition vectorized
+        duration_mask = self._screen_sample_times[possible_indices + chunk_size - 1] < self.end_time
+
+        # this assumes that the valid_condition is a single condition
+        valid_conditions = [self.modality_config["screen"]["valid_condition"]]
+
+        if self.modality_config["screen"]["include_blanks"]:
+            additional_valid_conditions = {"tier": "blank"}
```
pollytur, line 596: why blank is a `tier` now?

PR #72, opened 2025-05-20
Automated Formatting and Screeninterpolation testing

tomschewski:
- Added github workflow that automatically formats code into black standard and sorts imports.
- Added test for screen interpolation and got sequence tests running again
- Fixed merge issue in linear interpolation (Not very clean but will get revised anyway when seperating sequence interpolators)
- Removed deprecated warning from using ToTensor
PR seems very large but most of the changes are just one time formatting adjustments made by the test.


experanto/datasets.py
```diff
@@ -460,83 +487,112 @@ def initialize_transforms(self):
...
+                transform_list = [
+                    v
+                    for v in self.modality_config.screen.transforms.values()
+                    if isinstance(v, torch.nn.Module)
+                ]
                 transform_list.insert(0, add_channel)
             else:
-                transform_list = [ToTensor()]
+                transform_list = [
+                    Compose([ToImage(), ToDtype(torch.float32, scale=True)])
```
reneburghardt, line 498: Summary of a discussion between @tomschewski and myself:
Although `v2.Compose([v2.ToImage(), v2.ToDtype(torch.float32, scale=True)])` is the recommended way to get rid of the deprecated `ToTensor`, we don't think its ideal. Neither `ToImage()` in the case of `device_name != 'screen'` nor the hardcoded `torch.float32` nor the `scale=True` looks right here.
We assume `ToTensor()` was used here to transform the data from numpy format to a torch tensor mainly, which would lead to a solution using `torch.from_numpy` or similar.

tests/create_screen_data.py
```diff
@@ -0,0 +1,79 @@
...
+    duration=10.0, frame_shape=(32, 32), fps=10.0, image_frame_count=10, num_videos=1
+):
+
+    try:
+        data_dir = SCREEN_ROOT / "data"
+        meta_dir = SCREEN_ROOT / "meta"
+        data_dir.mkdir(parents=True, exist_ok=True)
+        meta_dir.mkdir(parents=True, exist_ok=True)
+
+        n_frames = int(np.round(duration * fps))
```
reneburghardt, line 22: Petty comment, but shouldn't it be np.floor? In case of 1,5s with fps of 5, this gives me 7,5, rounded to 8, although I would expect 7, no?

experanto/datasets.py
pollytur: @tomschewski could we please replace it with `torch.from_numpy` (because some of the variables are 1-d vectors and ToImage() is definitely wrong for it)

this would also solve the dtype but still - @reneburghardt just for your info - dtype its currently hardcoded in many places after the FM merge, and we should fix it later, ideally it should be configurable with a yaml file to

tomschewski: This is what I think is the most fitting replacement.
```
                transform_list = Compose([
                    lambda arr: torch.from_numpy(arr.transpose((2, 0, 1))).contiguous()
                ])

```

experanto/interpolators.py
```diff
@@ -187,46 +205,51 @@ def interpolate(self, times: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
...
+                interpolated = np.full(
+                    (valid_times.shape[0], data_lower.shape[1]), np.nan
+                )
+                interpolated[compute_mask] = (
+                    lower_signal_ratio * data_lower + upper_signal_ratio * data_upper
+                )
 
-                valid_indices = np.flatnonzero(valid)
-                valid[valid_indices[overflow_mask]] = False
+                # valid_indices = np.flatnonzero(valid)
```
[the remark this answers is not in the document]
tomschewski: Because there were major changes to the valid argument from what I can tell. It's now a list containing indices rather than a bool mask. Hence old adjustment of it does not work anymore. I adjusted it to work with the new ones in a new commit.

tests/create_screen_data.py
tomschewski: Yes.

experanto/interpolators.py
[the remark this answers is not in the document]
pollytur: It because before Konsti's FM PR the valid used to be a boolean mask ([link](https://github.com/sensorium-competition/experanto/blob/b25e5e0284d451c69965b8fd783ba5fef7683587/experanto/interpolators.py#L25-L26)) but now its indexes of the points to be used ([link](https://github.com/sensorium-competition/experanto/blob/main/experanto/intervals.py#L22-L23), basically same but with `np.where` around).
Tbh - I am not sure why we return the `valid` at all (see issue #73 ) - but I want to discuss it with Fabian and Kosnti before removing it
@tomschewski could we also please remove the commented out lines from the PR completely :)

experanto/interpolators.py
```diff
@@ -187,46 +205,51 @@ def interpolate(self, times: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
...
+                    )
 
-                    interpolated[dim_mask, dim] = lower_signal_ratio * data_lower + upper_signal_ratio * data_upper
-
-                valid_indices = np.flatnonzero(valid)
-                for mask in overflow_mask.T:
-                    valid[valid_indices[mask]] = False
+                # valid_indices = np.flatnonzero(valid)
+                # for mask in overflow_mask.T:
+                # valid[valid_indices[mask]] = False
```
pollytur, line 222:
```suggestion
```

experanto/interpolators.py
```diff
@@ -187,46 +205,51 @@ def interpolate(self, times: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
...
+                    (valid_times.shape[0], data_lower.shape[1]), np.nan
+                )
+                interpolated[compute_mask] = (
+                    lower_signal_ratio * data_lower + upper_signal_ratio * data_upper
+                )
 
-                valid_indices = np.flatnonzero(valid)
-                valid[valid_indices[overflow_mask]] = False
+                # valid_indices = np.flatnonzero(valid)
+                # valid[valid_indices[overflow_mask]] = False
```
pollytur, line 251:
```suggestion
```

.github/workflows/test.yml
```diff
@@ -9,23 +9,63 @@ on:
...
+          python-version: '3.12.8'
+
+      - name: Install tools
+        run: |
+          python -m pip install --upgrade pip
+          pip install "black[jupyter]" isort
+
+      - name: Run Black and Isort
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" && "${{ github.event.pull_request.head.repo.full_name }}" == "${{ github.repository }}" ]]; then
```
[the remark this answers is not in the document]
tomschewski: Changed as discussed.

experanto/datasets.py
```diff
@@ -460,83 +487,117 @@ def initialize_transforms(self):
...
+                    if isinstance(v, torch.nn.Module)
+                ]
                 transform_list.insert(0, add_channel)
             else:
-                transform_list = [ToTensor()]
+
+                transform_list = Compose(
+                    [
+                        lambda arr: torch.from_numpy(
+                            arr.transpose((2, 0, 1))
```
pollytur, line 501: @tomschewski why do you transpose here?

pollytur: 1) the first link to the doc is incorrect as its `torchvision.transforms.ToTensor` to while we used to import `from torchvision.transforms.v2 import ToTensor` ([this](https://docs.pytorch.org/vision/main/_modules/torchvision/transforms/v2/_deprecated.html#ToTensor) is the correct but second one is basically same)

However, this would fail as this is applied when modality is not! screen, hence, could be 1-2 dimensional!
in the original code they did 
```
if pic.ndim == 2:
    pic = pic[:, :, None]
```
([link](https://docs.pytorch.org/vision/main/_modules/torchvision/transforms/functional.html))
2) BUT I don't think there should be anything more but just  `lambda arr: torch.from_numpy(arr).contiguous()`
because again, its not necesarily an image here and I do not see a reason to overcomplicate it here

experanto/datasets.py
```diff
@@ -460,83 +487,117 @@ def initialize_transforms(self):
...
+                        ).contiguous()
+                    ]
+                )
```
pollytur, line 504:
```suggestion
                transform_list = Compose(
                    [
                        lambda arr: torch.from_numpy(
                            arr
                        ).contiguous()
                    ]
                )
```

experanto/datasets.py
pollytur: @tomschewski as replacing ToTensor if not that straightforward, I roll it back and lets take a look at this later

experanto/datasets.py
```diff
@@ -1,28 +1,30 @@
...
+from hydra.utils import instantiate
+from omegaconf import DictConfig, ListConfig, OmegaConf
 from torch.utils.data import Dataset
 from torchvision.transforms import v2
-from torchvision.transforms.v2 import ToTensor, Compose, Lambda
-from omegaconf import OmegaConf, DictConfig, ListConfig
-from hydra.utils import instantiate
-import functools
-import importlib
+from torchvision.transforms.v2 import Compose, Lambda, ToDtype, ToImage
```
pollytur, line 19:
```suggestion
from torchvision.transforms.v2 import Compose, Lambda, ToTensor
```

experanto/datasets.py
pollytur, line 504:
```suggestion

                 transform_list = [ToTensor()]
```

Issue #64, opened 2025-05-12
Decouple phase shifts from SequenceInterpolator into extended own Interpolator


reneburghardt:
There are a lot of if statements whether there are phase shifts or not in the sequence interpolator. An idea might be to implement an own interpolator extending the sequence interpolator.

reneburghardt:
Merged: https://github.com/sensorium-competition/experanto/pull/74

PR #74, opened 2025-05-22
Phase shift sequence interpolators

reneburghardt:
https://github.com/sensorium-competition/experanto/issues/64

- separate sequence interpolator with phase shifts into an own interpolator class
- simplified linear interpolation
- expand tests
- introduce close function to delete/close open resources (tests failed otherwise on my hpc setup)
- run tests verbosely on github


.github/workflows/test.yml
pollytur, line 1: could we fix the version of black, isort and pip in the installations here?

experanto/interpolators.py
```diff
@@ -39,18 +37,40 @@ def interpolate(self, times: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
...
+            )
 
     def valid_times(self, times: np.ndarray) -> np.ndarray:
         return self.valid_interval.intersect(times)
 
+    def close(self):
+        ...
+        # generally, nothing to do
+        # can be overwritten to close any open files or resources
+
```
pollytur, line 73: why do we need this? 
like its not called anywhere in practice and I do not see why would it but I might be missing sth
(sorry for being so picky, I just try that we do not overload experanto with software for the sake of software)

experanto/interpolators.py
```diff
@@ -83,33 +99,26 @@ def __init__(
         self.time_delta = 1.0 / self.sampling_rate
         self.start_time = meta["start_time"]
         self.end_time = meta["end_time"]
+        self.is_mem_mapped = meta["is_mem_mapped"] if "is_mem_mapped" in meta else False
```
pollytur, line 102: TODO - polly check meta files from the last export that they have this flag

.github/workflows/test.yml
reneburghardt: Do you want it to be here or in another separate PR?

experanto/interpolators.py
```diff
@@ -140,113 +149,154 @@ def normalize_data(self, data):
         return data
 
     def interpolate(self, times: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
-
         valid = self.valid_times(times)
         valid_times = times[valid]
+
+        if len(valid_times) == 0:
+            return np.array([]), valid
```
pollytur, line 156: Do we want to add a warning here that no points are valid for the queried time?
like 

```suggestion
        if len(valid_times) == 0:
            warnings.warn('Sequence interpolation returns empty array, no valid times queried')
            return np.array([]), valid
```
+ `import warnings` in the beginning

experanto/interpolators.py
pollytur: ok, I see, thanks a lot for the explanation.
I guess it's another case of python garbage collector not being efficient enough. Let me finish going through the tests and decide if we keep it or not:)

tests/test_sequence_interpolator.py
```diff
@@ -0,0 +1,429 @@
...
+        ), "Interpolation object is a PhaseShiftedSequenceInterpolator"
+
+        times = timestamps[:10] + 1e-9
+        interp, valid = seq_interp.interpolate(
+            times=times
+        )  # Add a small epsilon to avoid floating point errors
+        assert times.shape == valid.shape, "All samples should be valid"
+        assert np.allclose(
+            interp, data[:10]
+        ), "Nearest neighbor interpolation does not match expected data"
```
pollytur, line 38: I think here it should not be `allclose` but exact comparison as the values are not actually interpolated but take NN
```suggestion
        assert (interp == data[:10]).all(), "Nearest neighbor interpolation does not match expected data"
```

tests/test_sequence_interpolator.py
pollytur, line 1: I am a bit concerned by the `magic` 10, which is through this whole file. Can we make it a global variable like `SEQUENCE_LENGTH = 10` at the top of the file and just cause it everywhere in the tests, please?

tests/test_sequence_interpolator.py
```diff
@@ -0,0 +1,429 @@
...
+            expected_valid == valid
+        ).all(), "Valid times does not match expected values"
+        assert interp.shape == (
+            3,
+            10,
+        ), f"Expected interp.shape == (2, 10), got {interp.shape}"
+
+
+@pytest.mark.parametrize("interpolation_mode", ["nearest_neighbor", "linear"])
+@pytest.mark.parametrize("end_time", [0.2, 1.0, 5.0])
```
pollytur, line 366: Could we add the test case when the valid should be empty? And the end time above the queried times,
like
```suggestion
@pytest.mark.parametrize("end_time", [0.091, 0.2, 1.0, 5.0, 12])
```
Also, what is the difference between 0.2 and 1.0 test cases? both should have length 1, or am I missing sth? (or is it exactly the point?)

tests/test_sequence_interpolator.py
```diff
@@ -0,0 +1,429 @@
...
+        assert isinstance(
+            seq_interp, PhaseShiftedSequenceInterpolator
+        ), "Interpolation object is not a PhaseShiftedSequenceInterpolator"
+        seq_interp.interpolation_mode = interpolation_mode
+
+        times = np.array([-5.0, -0.1, 0.1, 4.9, 4.9999999, 5.0, 5.0000001, 5.1, 10.0])
+        interp, valid = seq_interp.interpolate(times=times)
+        assert (
+            np.where((times >= 0.09) & (times <= end_time + 0.09))[0] == valid
+        ).all(), "Valid times does not match expected values"
```
pollytur, line 390: why do we add sth for the `end_time` here?
We do not do it like this in the intervals.intersect ([link](https://github.com/sensorium-competition/experanto/blob/main/experanto/intervals.py#L27-L28))
```
def intersect(self, times: np.ndarray) -> np.ndarray:
        return np.where((times >= self.start) & (times <= self.end))[0]
```
and if 0.09 is start_time - could we give it a name and then use:) (for the sake of clarity)
```suggestion
        start_time = 0.09
        assert (
            np.where((times >= start_time) & (times <= end_time))[0] == valid
        ).all(), "Valid times does not match expected values"
```

tests/create_sequence_data.py
```diff
@@ -15,6 +17,7 @@ def create_sequence_data(
     use_mem_mapped=False,
     t_end=10.0,
     sampling_rate=10.0,
+    contain_nans=False,
```
pollytur, line 20: future note - add a test case when start time is non-zero (might cause numerics issues)

pollytur: + add irregular option for timestamps

experanto/interpolators.py
pollytur: ok, lets keep the closing as I do not now the better elegant solution for the tests right now

tests/create_sequence_data.py
```diff
@@ -47,6 +57,7 @@ def create_sequence_data(
             fp = np.memmap(filename, dtype=data.dtype, mode="w+", shape=data.shape)
             fp[:] = data[:]
             fp.flush()  # Ensure data is written to disk
+            del fp
         meta["dtype"] = str(data.dtype)
 
         if shifts_per_signal:
```
pollytur, line 63: @reneburghardt @tomschewski could some of you please add for the future backlog - to create the tests where the phase shifts are above the target sampling rate (now they are all smaller)

tests/test_sequence_interpolator.py
```diff
@@ -0,0 +1,429 @@
...
+        )
+    ) as (timestamps, data, _, seq_interp):
+        assert isinstance(
+            seq_interp, SequenceInterpolator
+        ), "Interpolation object is not a SequenceInterpolator"
+        assert not isinstance(
+            seq_interp, PhaseShiftedSequenceInterpolator
+        ), "Interpolation object is a PhaseShiftedSequenceInterpolator"
+
+        times = timestamps[:10] + 1e-9
```
pollytur, line 31: @reneburghardt tbh, I am not sure if this epsilon is needed here. Have you tried without it?

tests/test_sequence_interpolator.py
```diff
@@ -0,0 +1,429 @@
...
+
+        delta_t = 1.0 / sampling_rate
+        times = (
+            timestamps[1:11] + 1e-9
+        )  # Add a small epsilon to avoid floating point errors
+        interp, valid = seq_interp.interpolate(times=times)
+        assert times.shape == valid.shape, "All samples should be valid"
+        assert np.allclose(
+            interp, data[0:10]
+        ), "Nearest neighbor interpolation does not match expected data"
```
pollytur, line 139: again here - for knn interpolation it should be exact equality and not allclose

plus this test basically says that we need more generations for phase_shifts as the whole point is that they change the indexes of the data returned (eg for neuron 1 we return indexes 0 to 10 and for neuron 2 we return indexes from 1 to 11 because neuron 1 has a bigger timeshift than neuron 2)

But I am fine to add these tests later (@tomschewski - the explanation above might be also helpful for you )

tests/test_sequence_interpolator.py
```diff
@@ -0,0 +1,429 @@
...
+            n_signals,
+        ), f"Expected interp.shape == (10, {n_signals}), got {interp.shape}"
+
+        # Test phase shifts
+        for i in range(data.shape[1]):
+            for dt in np.linspace(0, 0.99) * delta_t:
+                shifted_times = times + shift[i] + dt
+
+                interp, valid = seq_interp.interpolate(times=shifted_times)
+                assert np.allclose(
```
pollytur, line 152: exact here instead of `allclose`

tests/test_sequence_interpolator.py
```diff
@@ -0,0 +1,429 @@
...
+                    interp[:, i], data[1:11, i]
+                ), f"Data at {dt} does not match original data (use_mem_mapped={use_mem_mapped}, sampling_rate={sampling_rate}, shifts_per_signal={True})"
+
+            for dt in np.linspace(1.0, 1.99) * delta_t:
+                shifted_times = times + shift[i] + dt
+
+                interp, valid = seq_interp.interpolate(times=shifted_times)
+                assert np.allclose(
+                    interp[:, i], data[2:12, i]
+                ), f"Data at {dt} does not match original data (use_mem_mapped={use_mem_mapped}, sampling_rate={sampling_rate}, shifts_per_signal={True})"
```
pollytur, line 162: exact here instead of `allclose`

tests/test_sequence_interpolator.py
```diff
@@ -0,0 +1,429 @@
...
+        ), "Interpolation object is not a PhaseShiftedSequenceInterpolator"
+
+        times = (
+            timestamps[1:11] + 1e-9
+        )  # Add a small epsilon to avoid floating point errors
+        interp, valid = seq_interp.interpolate(times=times)
+        assert times.shape == valid.shape, "All samples should be valid"
+        assert np.allclose(
+            interp, data[0:10], equal_nan=True
+        ), "Nearest neighbor interpolation does not match expected data"
```
pollytur, line 192: exact here instead of `allclose` (but just == would not work here because of nans, I guess)

pollytur, line 192: exact here instead of `allclose` (but just == would not work here because of nans, I guess)

tests/test_sequence_interpolator.py
```diff
@@ -0,0 +1,429 @@
...
+        assert valid.shape == (10,), f"Expected valid.shape == (10,), got {valid.shape}"
+        assert interp.shape == (
+            10,
+            n_signals,
+        ), f"Expected interp.shape == (10, {n_signals}), got {interp.shape}"
+
+
+@pytest.mark.parametrize("n_signals", [0, 1, 10, 50])
+@pytest.mark.parametrize("phase_shifts", [True, False])
+def test_linear_interpolation_replaces_nans(n_signals, phase_shifts):
```
pollytur, line 245: `expected` and comparison with it is missing in this test

experanto/interpolators.py
```diff
@@ -140,113 +149,154 @@ def normalize_data(self, data):
...
+            _, cols = np.indices(idx_lower.shape)
+            data_lower = self._data[idx_lower, cols]
+            data_upper = self._data[idx_upper, cols]
+
+            interpolated = (
+                lower_signal_ratio * data_lower + upper_signal_ratio * data_upper
+            )
 
-                valid = valid[~overflow_mask]
+            valid = valid[~overflow_mask.any(axis=1)]
```
pollytur, line 299: why is it `.any(axis=1)` here?

pollytur:
one conceptual comment wrt this and previous PR - we probably want not just replace nans with the mean values but sometimes actually interpolate nans linearly, within certain window (like if there are no more then 3 nan values in a row - they should be interpolated linearly) and only if the nans overflow the tolerance window - replace it with local means

I do not think this should be a part of this PR, rather a note for future

.github/workflows/test.yml
pollytur: In case you applied `isort` locally - i guess the reason the workflow failed is the versions of isort - because black clearly passed based on the logs; isort is after black it gives the errors

tests/test_sequence_interpolator.py
reneburghardt: What are the concerns? 

Replaced all 10 by a variable.

tests/test_sequence_interpolator.py
reneburghardt: Adjusted test cases regarding end_time

tests/test_sequence_interpolator.py
reneburghardt: For the phase shifts. They are not added in `intervals.intersect`, but here: https://github.com/sensorium-competition/experanto/blob/cac5668bf23d8c5969969e9a5ffb795788b593e5/experanto/interpolators.py#L242

I made it clearer.

experanto/interpolators.py
reneburghardt: When we have phase shifts, overflow can happen for some neurons (with a high phase shift) and not for others (with a low phase shift). So we check if an overflow is happening for any neuron at a certain time and disregard the time then completely.

When we don't have phase shifts, the times are the same for every neuron, leading to one dimension less and therefore `.any(axis=1)` is not necessary.

reneburghardt:
> we probably want not just replace nans with the mean values but sometimes actually interpolate nans linearly, within certain window (like if there are no more then 3 nan values in a row - they should be interpolated linearly) and only if the nans overflow the tolerance window - replace it with local means

Added an issue for that: https://github.com/sensorium-competition/experanto/issues/75

experanto/interpolators.py
pollytur: @reneburghardt  should we resolve this one as well?

tests/create_sequence_data.py
```diff
@@ -63,11 +63,13 @@ def create_sequence_data(
...
             shifts = np.random.rand(n_signals) / meta["sampling_rate"] * 0.9
             np.save(SEQUENCE_ROOT / "meta" / "phase_shifts.npy", shifts)
+        else:
+            shifts = np.zeros(n_signals)
 
         with open(SEQUENCE_ROOT / "meta.yml", "w") as f:
             yaml.dump(meta, f)
 
-        yield timestamps, data, shifts if shifts_per_signal else None
+        yield timestamps, data, shifts
```
pollytur, line 72: @reneburghardt why do we need to return zeros for the phase shifts now?

tests/test_sequence_interpolator.py
pollytur: @reneburghardt should we resolve this one as well, it seems to be fixed now, isn't it?

experanto/interpolators.py
pollytur: could we please keep this conversation open for now, I need to think a bit about it

tests/create_sequence_data.py
reneburghardt: Whoops, that remained there accidentally from a change I didn't do in the end. I reverted it and pushed :)

experanto/interpolators.py
pollytur: it is there for all the Sequence modalities indeed (but not for the `screen`)

PR #79, opened 2025-05-26
numerics for knn interpolation and minor clean up

pollytur:
* implements response normalization as in neural predictors (remove hardcoded threshold=1)
* minor code cleanup (remove non-used code)
* fix numerical issues when querying timepoints (truncate times to the 5th digit after the dot. Due to big value of start time in real datasets we cannot just cast the array to float32 but actually need float64)
* removes incorrect times for responses in __getitem__ (as interpolation implies that the times are same for all the neurons)

### DIFF
--- a/experanto/datasets.py
+++ b/experanto/datasets.py
@@ -257,9 +257,7 @@ class ChunkDataset(Dataset):
                     # https://github.com/sinzlab/neuralpredictors/blob/2b420058b2c0c029842ba739829114ddfa0f8b50/neuralpredictors/data/transforms.py#L375-L378
                     threshold = 0.01 * np.nanmean(stds)
                     idx = stds[0, :] < threshold  # response std shape: (1, n_neurons)
-                    stds[0, idx] = (
-                        threshold  # setting stds which are smaller than 1 to threshold
-                    )
+                    stds[0, idx] = threshold  # setting stds which are smaller than threshold to threshold
 
                 # if mode is a dict, it will override the means and stds
                 if not isinstance(mode, str):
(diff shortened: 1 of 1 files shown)


experanto/datasets.py
```diff
@@ -438,8 +253,13 @@ def initialize_statistics(self) -> None:
...
                 if device_name == "responses":
-                    idx = stds[0, :] < 1  # response std shape: (1, n_neurons)
-                    stds[0, idx] = 1  # setting stds which are smaller than 1 to 1
+                    # same as in neuralpredictors before
+                    # https://github.com/sinzlab/neuralpredictors/blob/2b420058b2c0c029842ba739829114ddfa0f8b50/neuralpredictors/data/transforms.py#L375-L378
+                    threshold = 0.01 * np.nanmean(stds)
+                    idx = stds[0, :] < threshold  # response std shape: (1, n_neurons)
+                    stds[0, idx] = (
+                        threshold  # setting stds which are smaller than 1 to threshold
+                    )
```
reneburghardt, line 262: Why the parenthesis?
And the comment should be the following, no?
```suggestion
                    stds[0, idx] = threshold  # setting stds which are smaller than threshold to threshold
```

experanto/datasets.py
```diff
@@ -802,8 +622,19 @@ def __getitem__(self, idx) -> dict:
             chunk_size = self.chunk_sizes[device_name]
             chunk_s = chunk_size / sampling_rate
 
-            times = np.linspace(s, s + chunk_s, chunk_size, endpoint=False)
-            times = times + self.modality_config[device_name].offset
+            # convert everything to int to avoid numerical issues
+            start_time = int(round(s * self.scale_precision))
+            offset = int(
+                round(
+                    sample_dataset.modality_config[device_name].offset
```
reneburghardt, line 629: I am so confused right now, how does this work? 😅 Where is `sample_dataset` defined?
And in general, does it make sense to first check if offset exists for the device config and if not, set offset to 0?

experanto/datasets.py
pollytur: parenthesis are coming from black due to the very long comments, so they will be back. But the comment should be updated indeed, thanks!

PR #81, opened 2025-06-02
Data compression for screen data

schewskone:
Implemented support for compressed screen formats

- Added EncodedImageTrial and EncodedVideoTrial which feature `get_data_() ` methods that support encoded formats.
- Extended screen tests and screen data generation to feature encoded data in mp4 and jpeg format.
- Thinned requirements.txt and added ffmpeg installation to CI workflow.

Right now the VideoDecoder decodes the entire video and returns that to the interpolation function. This can be optimized with the help of sliced decoding which would require changes to the interpolation method. @pollytur  and I decided we  should discuss  first and implement it later on if deemed necessary.

Issue #73, opened 2025-05-22
Remove returning `valid` for interpolation


pollytur:
Currently, we return `interpolated, valid` after interpolation in both `Interpolators` (`SequenceIterpolator` [here](https://github.com/sensorium-competition/experanto/blob/main/experanto/interpolators.py#L235) and [here](https://github.com/sensorium-competition/experanto/blob/main/experanto/interpolators.py#L162), `ScreenInterpolator` [here](https://github.com/sensorium-competition/experanto/blob/main/experanto/interpolators.py#L393)) and `Experiment` ([here](https://github.com/sensorium-competition/experanto/blob/main/experanto/experiment.py#L67)), however, its not used in the actual dataset (example for `ChunkDataset` [here](https://github.com/sensorium-competition/experanto/blob/main/experanto/datasets.py#L719)) and I do not see a case where we would need this data, hence I suggest to discuss this and maybe remove `valid` and return only the interpolated values.

PR #84, opened 2025-07-03
added TimeIntervalInterpolator and tests for it.

oliveira-caio:
Initial implementation of an interpolator for a new type of modality that, for now, we're calling "TimeInterval". The idea for this new modality is that it's neither sequence nor events nor screen, but actually some sort of discrete data type that we want to interpolate. Use cases:

- Tier separation: we usually have a list of intervals for splitting the data into training, testing and validation. We want to be able to query for a certain time point and figure out to which tier that time point belongs to.
- Trial property: in my particular case we have trials and in each trial the monkey can turn left or right. I want to be able to know which side the monkey chose in a given time point.

I also implemented a very basic test that I recommend the reviewer to take a look to check if it makes sense.

### DIFF
--- a/experanto/interpolators.py
+++ b/experanto/interpolators.py
@@ -530,7 +530,7 @@ class TimeIntervalInterpolator(Interpolator):
         - Only time points within the valid interval (as defined by start_time and
           end_time in meta.yml) are considered; others are filtered out.
         - Intervals where start > end are considered invalid and will trigger a
-          warning. Zero-length intervals (start == end) are skipped.
+          warning.
         """
         valid = self.valid_times(times)
         valid_times = times[valid]


.gitignore
oliveira-caio, line 1: i added this because i use a symlink to a datasets folder that i share across projects and for some reason symlinks don't add the `/` add the end.

reneburghardt:
@oliveira-caio Can you please run black and isort on your local repo and commit and push changes in formatting then? The changes of this PR seem to entail some formatting-only changes. Make sure to run black and isort with the following versions to ensure compatibility with github actions: `pip install "black[jupyter]==25.1.0" isort==6.0.1`

experanto/interpolators.py
```diff
@@ -478,6 +479,64 @@ def rescale_frame(self, frame: np.array) -> np.array:
...
+            else:
+                label_intervals = np.load(
+                    self.root_folder / f"{label}.npy", allow_pickle=True
+                )
+
+            for start, end in label_intervals:
+                mask = (valid_times >= start) & (valid_times < end)
+                out[mask, l] = True
+
+        return out, valid
```
pollytur, line 537: could we remove returning `valid` from here please? 
This is in #73 and @schewskone is working on it for other interpolators in #81

experanto/interpolators.py
```diff
@@ -478,6 +479,64 @@ def rescale_frame(self, frame: np.array) -> np.array:
...
+        self.cache_data = cache_data
+
+        meta = self.load_meta()
+        self.labels = list(sorted(meta["labels"].keys()))
+        self.start_time = meta["start_time"]
+        self.end_time = meta["end_time"]
+        self.valid_interval = TimeInterval(self.start_time, self.end_time)
+
+        if self.cache_data:
+            self.label_data = {
```
pollytur, line 494: opinionated nitpick (e.g. feel free to ignore) but I would call it 
```suggestion
            self.labeled_interval = {
```
as they are intervals while data could anything

tests/create_time_intervals_data.py
```diff
@@ -0,0 +1,44 @@
...
+from pathlib import Path
+
+import numpy as np
+import yaml
+
+TIME_INTERVAL_ROOT = Path("tests/time_interval_data")
+
+
+@contextmanager
+def create_time_intervals_data():
```
pollytur, line 12: this is a nice start but I think we need a couple of more test cases where
1. the intervals are irregular (right now everything is in steps of 200)
2. the intervals are overlapping 
    * 2 overlapping together 
    * 3 overlapping together
    * a break time in between (eg not train-test-val between 450 and 500)
    
@oliveira-caio could we please add these?

pollytur (changes requested):
thanks a lot, overall looks good

I guess we need a bit more tests, fix loading files and remove returning  valid from `interpolate` function

@oliveira-caio please let me know when the fixes are here, I would test it with the actual mocked data locally afterwards:)

.gitignore
```diff
@@ -2,7 +2,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
-data/
+data*
```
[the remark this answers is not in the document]
pollytur: thanks for the explanation:)
should be fine for now as we not have any `data` folder and the change makes sense, so its ok

experanto/interpolators.py
```diff
@@ -58,6 +58,8 @@ def create(root_folder: str, cache_data: bool = False, **kwargs) -> "Interpolato
                 return SequenceInterpolator(root_folder, cache_data, **kwargs)
         elif modality == "screen":
             return ScreenInterpolator(root_folder, cache_data, **kwargs)
+        elif modality == "time_interval":
+            return TimeIntervalInterpolator(root_folder, cache_data, **kwargs)
         else:
             raise ValueError(
                 f"There is no interpolator for {modality}. Please use 'sequence' or 'screen' as modality."
```
pollytur, line 65:
```suggestion
                f"There is no interpolator for {modality}. Please use 'sequence', 'screen' or 'time_interval' as modality."
```
inspired by copilot in another PR but belongs here

experanto/filters/common_filters.py
```diff
@@ -9,8 +9,7 @@
 
 
 def nan_filter(vicinity=0.05):
-
-    def implementation(device_: Interpolator):
+    def implementation(device_: SequenceInterpolator):
```
pollytur, line 12:
```suggestion
    def implementation(device_: SequenceInterpolator):
        # requests SequenceInterpolator as uses time_delta internally
        # and other interpolators don't have it
```

experanto/interpolators.py
```diff
@@ -478,6 +479,62 @@ def rescale_frame(self, frame: np.array) -> np.array:
...
+
+        meta = self.load_meta()
+        self.meta_labels = meta["labels"]
+        self.start_time = meta["start_time"]
+        self.end_time = meta["end_time"]
+        self.valid_interval = TimeInterval(self.start_time, self.end_time)
+
+        if self.cache_data:
+            self.labeled_intervals = {
+                label: np.load(filename, allow_pickle=True)
```
pollytur, line 495: and why do you still want to use `allow_pickle=True`?

pollytur: also, should it be f`{self.root_folder}/filename`, shouldn't it?

tests/create_time_intervals_data.py
```diff
@@ -0,0 +1,96 @@
...
+            np.save(TIME_INTERVAL_ROOT / "test.npy", np.array([[0, 150]]))
+            np.save(
+                TIME_INTERVAL_ROOT / "train.npy",
+                np.array([[200, 400], [600, 800]]),
+            )
+            np.save(
+                TIME_INTERVAL_ROOT / "validation.npy",
+                np.array([[400, 600], [800, t_end]]),
+            )
+            timestamps[100:250] = np.nan
```
pollytur, line 91: what would this test case mean? 
I am not sure if I understand what is supposed to happen

tests/test_time_intervals_interpolator.py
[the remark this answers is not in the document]
oliveira-caio: if start = end, do we want all values to be False or should the value at start = end be True?

experanto/interpolators.py
```diff
@@ -478,6 +479,81 @@ def rescale_frame(self, frame: np.array) -> np.array:
...
+                        f"TimeIntervalInterpolator found no intervals for label: {label}"
+                    )
+                    continue
+
+            for start, end in intervals:
+                if start >= end:
+                    warnings.warn(
+                        f"Invalid interval found for label: {label}, interval: ({start}, {end})"
+                    )
+                    continue
```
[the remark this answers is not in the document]
oliveira-caio: To me it's a little vague what copilot means by "suggest" when it says "the test cases suggest they should be valid". I double checked the tests and didn't spot anything wrong. One thing, though: the conditional `if start >= end` is only necessary to warn the user, because the logic of the intersection of the valid time points is correct even if start >= end (everything will be False in this case, as expected). So, the question is: when do we warn the user that the interval is invalid: when start > end or when start >= end? It's seems like a minor issue though.

PR #85, opened 2025-07-04
Querying by time and move data interpolating into experiment

reneburghardt:
**Still needs testing**


experanto/experiment.py
```diff
@@ -72,3 +79,164 @@ def interpolate(self, times: slice, device=None) -> tuple[np.ndarray, np.ndarray
 
     def get_valid_range(self, device_name) -> tuple:
         return tuple(self.devices[device_name].valid_interval)
+
+    def get_data_for_interval(
```
schewskone: Is this function only for testing purposes? I did not see it anywhere but the tests. Do we need this?

tests/create_experiment.py
```diff
@@ -0,0 +1,53 @@
...
+    default_devices_kwargs = {
+        "sampling_rate": 1.0,
+    }
+    devices_kwargs = [default_devices_kwargs | kwargs for kwargs in devices_kwargs]
+    try:
+        EXPERIMENT_ROOT.mkdir(parents=True, exist_ok=True)
+
+        for device_id, device_kwargs in enumerate(devices_kwargs):
+            device_path = EXPERIMENT_ROOT / f"device_{device_id}"
+            _generate_sequence_data(device_path, **device_kwargs)
```
schewskone: Why do only we instantiate two sequence devices and no screen? We would have to change `create_screen_data.py` for this as well right?

experanto/experiment.py
schewskone: Ok good to know. Thanks

schewskone (approved):

pollytur:
@reneburghardt why have you decided to move chunks to the Experiment level? What is the high level motivation behind this decision? (I think I missed this discussion)

reneburghardt:
> @reneburghardt why have you decided to move chunks to the Experiment level? What is the high level motivation behind this decision? (I think I missed this discussion)

To get data with regards to chunk size or with regards to interval seemed to me to be on the same level or, in other words, having both options not in the same place felt weid. They also share a fair bit of code such as scaling for precision and resolving sampling rate and offset.

How do you see that @pollytur ?

Issue #82, opened 2025-07-03
Assumption that is not satisfied


fabiansinz:
This refers to a `time_delta` which is - rightfully - not provided by the Interpolator. This should therefore request a `SequenceInterpolator`. 

https://github.com/sensorium-competition/experanto/blob/ff39929d54f2b2006a08c517c2758dd304ba5597/experanto/filters/common_filters.py#L13

PR #86, opened 2025-07-04
Fixed nan filter issue

oliveira-caio:
Fixed issue #82. There is a question, though: should we allow other interpolators other than sequences to have sampling rate and be filtered by nan? Screen doesn't have a sampling rate because it's a special case and I feel like the TimeIntervalInterpolator will be created out of the screen meta files at export time and, therefore, may also not have a sampling rate. So, at the moment the only type of interpolator that can be filtered by nan is the SequenceInterpolator, but we may have to change this in the future.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,3 +40,6 @@ skip = [".ipynb_checkpoints"]
 profile = "black"
 line_length = 88
 known_first_party = ["experanto"]
+
+[tool.ruff]
+exclude = ["**"]
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
-data/
+data*
 local dev/
 
 # C extensions
--- a/experanto/filters/common_filters.py
+++ b/experanto/filters/common_filters.py
@@ -1,6 +1,6 @@
 import numpy as np
 
-from experanto.interpolators import Interpolator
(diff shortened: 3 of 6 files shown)


pollytur:
@oliveira-caio this PR now has changes from #84 and its own changes, which is a bit of a mess
as #84 requires additional changes after the review - this is going to cause conflicts. Could we please rebase it? 
Sth like this should work
```
git checkout main
git pull origin main

git checkout pr2-branch
git rebase -i main
```

For future please base new branches off of main if there are interdependent parts, to avoid this:)

pyproject.toml
```diff
@@ -40,3 +40,6 @@ skip = [".ipynb_checkpoints"]
 profile = "black"
 line_length = 88
 known_first_party = ["experanto"]
+
+[tool.ruff]
+exclude = ["**"]
```
pollytur, line 45: and again why? do you have different setting than the isort and black versions we are using in here?

pollytur:
@oliveira-caio thanks for the PR.
As for Nans - lets keep it a sequenceInterpolator for now.
I am not sure if Gazes or other interval modalities could have nans tbh
For screen - ideally this should not be the case and I have seen such a thing, so I am not sure if we want to cover the non-existing cases but maybe worth talking with Fabian about it.
Anyway, I would say that's outside this PR :)

PR #97, opened 2026-01-27
Benchmarking Dataloaders

schewskone:
PR Draft, only contains a benchmarking function for the Multiepochsdataloader class.
Pretty straight forward but works nicely.

I also have benchmarks for the Interpolators and the Decoding / Numpy loading speeds but I'm not sure if we need that inside the experanto repo, it could be it's own script inside the repo that generates data or same as here a benchmark method for the classes but it's a bit more complicated to loop over the actual data.

### DIFF
--- a/experanto/utils.py
+++ b/experanto/utils.py
@@ -152,8 +152,10 @@ class MultiEpochsDataLoader(torch.utils.data.DataLoader):
                 print(f"Batch {i+1}: {batch_times[-1]:.4f}s")
 
         avg_time = sum(batch_times) / len(batch_times)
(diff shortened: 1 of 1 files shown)


experanto/utils.py
```diff
@@ -121,6 +121,42 @@ def __iter__(self):
         for i in range(len(self)):
             yield next(self.iterator)
 
+    def benchmark(
```
reneburghardt: IMO we shouldn't include any benchmarking functions into the actual package and especially not into a specific dataloader, but rather have scripts in a separate place to benchmark important parts of experanto and easily restart benchmarks after changes to see if runtime is affected

PR #99, opened 2026-02-27
Use yaml.safe_load/safe_dump for meta YAML

### LINKED ISSUE #94: Use yaml.safe_ functions
Replace yaml.dump and yaml.load with yaml.safe_dump and yaml.safe_load to avoid emitting and parsing Python-specific object tags. This improves security (prevents arbitrary object construction on load) and ensures YAML files remain portable and standards-compliant while it should still be working well for our use case as we only need standard Python objects in our meta files.

Vrittigyl:
### Summary
Replaced `yaml.load` / `yaml.dump` with `yaml.safe_load` / `yaml.safe_dump` in meta YAML handling.

### Reason
- Prevents arbitrary object construction and avoids Python-specific YAML tags.  
- Meta files only require standard Python types, so `safe_*` functions are sufficient and more secure.

### Verification
- `pytest` passes locally  

Closes #94

### DIFF
--- a/examples/sensorium/interpolator_demo.ipynb
+++ b/examples/sensorium/interpolator_demo.ipynb
@@ -57,7 +57,7 @@
     ")\n",
     "\n",
     "with open(\"/Users/fabee/Data/sinzlab-data/dataset0/eye_tracker/meta.yml\", \"w\") as f:\n",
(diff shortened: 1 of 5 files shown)


pollytur (approved):
LGTM

Issue #104, opened 2026-03-02
Add pycheck (or pyright)  and Codecov (for test-coverage reporting) to CI/CD


pollytur:

pollytur:
@reneburghardt  @schewskone any opinions if we should do it? 
I suggest to add codevoc to show test coverage in the PR trigger and pycheck to CI/CD as we have black and isort. 
It should not be too painful since I mostly already fixed pycheck in the last PR and from now on it should only be  maintain it when adding new staff

PR #106, opened 2026-03-02
Add pycheck and codecov

pollytur:
* adds pycheck and codevoc to the CI/Cd as suggested in #104

The codevoc token should be already setup in the current repo but it would not be fully active before this PR is actually merged

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,4 +48,4 @@ known_first_party = ["experanto"]
 [tool.pyright]
 include = ["experanto"]
 typeCheckingMode = "basic"
-pythonVersion = "3.9"
+pythonVersion = "3.12"


experanto/interpolators.py
```diff
@@ -453,8 +453,8 @@ def interpolate(
         valid_times += 1e-4  # add small offset to avoid numerical issues
 
         assert np.all(np.diff(valid_times) > 0), "Times must be sorted"
-        idx = (
-            np.searchsorted(self.timestamps, valid_times) - 1
+        idx = cast(
+            np.ndarray, np.searchsorted(self.timestamps, valid_times) - 1
```
reneburghardt: If we start using `cast` here, maybe its also better to do it in other places where you used type ignore comments (`# type: ignore ...`) as well? But not in this PR I reckon

pollytur: I'm not sure it would work there, here its about dimensions mostly but in the data loaders where I did the ignore things - it was a union with optional or sth like this -  not sure if casting would work there

pyproject.toml
```diff
@@ -40,3 +43,8 @@ skip = [".ipynb_checkpoints"]
 profile = "black"
 line_length = 88
 known_first_party = ["experanto"]
+
+[tool.pyright]
+include = ["experanto"]
+typeCheckingMode = "basic"
+pythonVersion = "3.9"
```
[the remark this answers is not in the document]
pollytur: @reneburghardt done :)

Issue #76, opened 2025-05-23
Expand test cases for sequence interpolator


reneburghardt:
From Polly:
a)
> add a test case when start time is non-zero (might cause numerics issues)

b)
> add irregular option for timestamps

c)
> where the phase shifts are above the target sampling rate (now they are all smaller)
PS for clarity - target sampling rate is the one we resample data later, not the one we generate it with

d)
> we need more generations for phase_shifts as the whole point is that they change the indexes of the data returned (eg for neuron 1 we return indexes 0 to 10 and for neuron 2 we return indexes from 1 to 11 because neuron 1 has a bigger timeshift than neuron 2)

e)
> where the expected would be computed with np.interp (maybe removed functions from utils might help)
this would basically test the assumptions on how we do it

see https://github.com/sensorium-competition/experanto/pull/74

BitForge95:
Hi @reneburghardt! This is incredibly helpful, thank you for sharing it. I will definitely take a look at your draft PR and see how I can build upon your create_experiment.py and test_experiment.py files. I will try to integrate those ideas with the initial test structure I have been working on and open a draft PR soon

pollytur:
a and b are addressed in #113

PR #113, opened 2026-03-04
Add test coverage for Experiment class and data generation utilities

BitForge95:
Fixes points a and b of #76

Hi @pollytur and @reneburghardt,

Following up on my progress in the issue thread, I’m opening this PR to add the test coverage for `experiment.py`. I've focused on making sure the `Experiment` class can handle the real-world data issues we discussed, like non-zero start times and irregular sampling.

### What I’ve done:

* **Experiment Class:** I updated the class to support three ways of getting an interpolator: Hydra, existing objects, or the original logic. I also added logic to track the global start and end times across all devices.
* **Data Generator:** I built `create_sequence_data.py` to create temporary test folders. This was key for testing specific problems without needing real files. I added a `start_time` parameter for testing offsets (like starting at 1.5s) and an `irregular` flag to simulate sensor jitter.
* **Environment Setup:** I used a context manager in `create_experiment.py` to handle the setup and teardown of the test folders. It makes sure everything is cleaned up after the tests run so the project directory stays tidy.

### Testing Strategy:

I split the tests in `test_experiment.py` into two parts:
1. **Routing logic:** I used mocks to make sure the `Experiment` class is sending data to the right interpolators.
2. **Integration:** I used the generator to verify that the code handles real file inputs and calculates valid ranges correctly.

To handle the numeric problems mentioned in the thread, I used `pytest.approx` for time comparisons and tested the code with non-integer sampling rates (like 33.33 Hz) to ensure the math stays stable.

Thanks again to @reneburghardt for the help with the initial structure!


BitForge95:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/sensorium-competition/experanto/pull/113#pullrequestreview-3889277359)

experanto/experiment.py
```diff
@@ -93,26 +84,22 @@ def interpolate(
         return_valid: bool = False,
     ) -> Union[tuple[dict, dict], dict, tuple[np.ndarray, np.ndarray], np.ndarray]:
```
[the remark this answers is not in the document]
pollytur: @BitForge95 please dont blindly throw AI on top of its feedback. Its not always right and sometimes - as here - the suggestion is not straightforward and requires a discussion / strategic decision

pollytur (changes requested):

experanto/experiment.py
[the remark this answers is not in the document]
BitForge95: Hi @pollytur, you are completely right, my apologies. I was too quick to apply the Copilot suggestion to make the type hint pass without fully considering the architectural design. Restricting it to Union[str, None] makes perfect sense so the Experiment strictly manages its own initialized devices. I am reverting that specific change to the interpolate method now and updating the type hint to match your suggestion. Thanks for catching this and for the guidance!

tests/create_experiment.py
```diff
@@ -0,0 +1,54 @@
+import copy
+import shutil
+from contextlib import contextmanager
+from pathlib import Path
+
+from .create_sequence_data import _generate_sequence_data
+
+# Temporary directory for test execution
+EXPERIMENT_ROOT = Path("tests/experiment")
+
```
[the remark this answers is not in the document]
BitForge95: @pollytur I don't think its suggested fix (tempfile) is the best fit. Since we are using pytest, I'd prefer to refactor create_experiment to accept the built-in tmp_path fixture. This ensures every test gets a perfectly isolated, automatically cleaned-up directory. Shall I go ahead and implement the tmp_path refactor?

tests/create_sequence_data.py
```diff
@@ -1,83 +1,124 @@
...
+def _generate_sequence_data(
+    sequence_root,
     n_signals=10,
     shifts_per_signal=False,
     use_mem_mapped=False,
+    start_time=0.0,
     t_end=10.0,
     sampling_rate=10.0,
     contain_nans=False,
+    irregular=False,
```
pollytur: @BitForge95 did not you plan to remove the `irregular` flag since is should not be unused?

BitForge95: Yes, I completely forgot to remove it when I deleted the test. I will just remove the flag and push the updated code, along with the other minor cleanups from the Copilot reviews.

experanto/experiment.py
```diff
@@ -77,47 +75,37 @@ def __init__(
...
 
             else:
-                # Default back to original logic
-                warnings.warn(
-                    "Falling back to original Interpolator creation logic.",
-                    UserWarning,
-                )
                 dev = Interpolator.create(
-                    d,
+                    str(d),
```
pollytur: why is it casted to be a string here now?

pollytur:
@BitForge95 could we please merge main into the current branch?

BitForge95:
@pollytur , I have merged the main , the code has been changed after i pushed the code. Thanks , i have pushed the latest code withour any merge conflicts

tests/test_experiment.py
```diff
@@ -1,12 +1,193 @@
...
+                assert isinstance(data, dict)
+                assert "device_0" in data and "device_1" in data
+                assert data["device_0"].shape == (2, 10)
+            else:
+                assert isinstance(data, np.ndarray)
+                assert data.shape == (2, 10)
 
-from .create_experiment import make_modality_config, make_sequence_device
 
 DEVICE_TIME_RANGE_CASES = [
```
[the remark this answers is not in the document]
BitForge95: I'll move these constants to the top of the file right after the imports

tests/test_experiment.py
pollytur: there is a test-case here missing when the union is actually a union and not just the longest device

```suggestion
# Three devices with different ranges: start should be min, end should be max
([(0.0, 10.0), (1.0, 8.0), (2.0, 9.0)], 0.0, 10.0),
([(1.0, 4.0), (3.0, 8.0), (2.0, 9.0)], 1.0, 9.0)
```

tests/test_experiment.py
```diff
@@ -1,12 +1,193 @@
...
+        experiment = Experiment(
+            root_folder=str(experiment_path),
+            modality_config=get_default_config(),
+        )
+        with pytest.raises(KeyError):
+            experiment.get_valid_range("device_does_not_exist")
+
+
+def test_experiment_with_non_zero_start_time(tmp_path):
+    """Test boundary conditions for data not starting at t=0."""
```
pollytur: I don't fully get this test and it purpose 
maybe lets discuss a bit

ideally we want to catch some correct error from interpolator if we query time outside of the valid region boundaries and some values if we are in the correct boundary but thats not what's happening here

pollytur (changes requested):
@BitForge95 thanks a lot!

Its now mostly comments about specific test case and less about style etc - please take a look :)

tests/test_experiment.py
```diff
@@ -1,12 +1,193 @@
...
+        # Confirm interpolation doesn't crash at the offset start
+        res = experiment.interpolate(np.array([start_offset]), device="device_0")
+        assert res is not None
+
+
+def test_experiment_numeric_precision_offset(tmp_path):
+    """Stress test using non-integer rates and offsets to catch float drift."""
+    start_offset = 0.123456789
+    sampling_rate = 33.3333333
+    duration = 1.0
```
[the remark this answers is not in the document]
pollytur: all the values need to be non-negative, sampling_rate should be strictly positive, otherwise not sure

not sure if we need a test case when sampling rate is smaller than duration (probably yes, not sure if we have it and if it should be in here)

pollytur (approved):
LGTM

@reneburghardt please take a look :)

pollytur:
@reneburghardt could you please take a look here when you have time?

pollytur:
@BitForge95 please merge current main in your branch :)

BitForge95:
Hi @pollytur , i have rebased the branch against the latest main and pushed everything into a single commit. Sorry it took me a bit too long to resolve the merge conflicts. But now its done. Let me know if I need to make any more changes :)

reneburghardt:
@BitForge95 I can see recent changes from e.g. https://github.com/sensorium-competition/experanto/pull/147 are reversed here (for example removing `workflow_dispatch` from all CI pipelines). Is that on purpose and if so, what is the reason?

experanto/interpolators.py
```diff
@@ -84,7 +84,9 @@ def __exit__(self, *exc):
         self.close()
 
     @staticmethod
-    def create(root_folder: str, cache_data: bool = False, **kwargs) -> Interpolator:
+    def create(
+        root_folder: str | Path, cache_data: bool = False, **kwargs
+    ) -> "Interpolator":
```
reneburghardt: Why do we add quotes around Interpolator here again? I thought its not necessary when we have `from __future__ import annotations`?

experanto/interpolators.py
```diff
@@ -109,35 +111,24 @@ def create(root_folder: str, cache_data: bool = False, **kwargs) -> Interpolator
...
         elif modality == "screen":
-            use_stimuli_names = kwargs.pop(
-                "use_stimuli_names", meta_data.get("use_stimuli_names", False)
-            )
-            return ScreenInterpolator(
-                root_folder,
-                cache_data=cache_data,
-                use_stimuli_names=use_stimuli_names,
-                **kwargs,
-            )
```
reneburghardt, line 134: Please don't remove recent changes from main here

BitForge95:
Hi @reneburghardt, this was definitely not on purpose. I recently did a rebase and adjusted my commits to clean up the PR history, and my local working directory was out of sync. As a result, I accidentally reverted some of the recent updates from main (like #147 and the interpolator changes). I am fixing this right now to restore all those files so this PR contains only the test coverage additions. Apologies for the confusion.

BitForge95:
@reneburghardt I have restored the CI workflows , pyproject.toml, and interpolators.py.

.readthedocs.yml
```diff
@@ -4,7 +4,7 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.12"
```
reneburghardt, line 7: @BitForge95 I think this also needs to be reverted

BitForge95: yeah sorry , I forgot.

BitForge95:
@reneburghardt i have revereted the changes.

pollytur:
@BitForge95 why dont we see the CI/CD checks anymore? also if I check it on your branch - ruff checks actually do not pass
@reneburghardt do you have an idea about CI / CD?

reneburghardt:
> @BitForge95 why dont we see the CI/CD checks anymore? also if I check it on your branch - ruff checks actually do not pass @reneburghardt do you have an idea about CI / CD?

Its always a bit weird when the last commit is from the auto-format bot, but beyond that I have no idea currently

BitForge95:
Hi @pollytur and @reneburghardt. The CI checks disappeared because the last commit on this branch was made automatically by the black/isort formatting bot. GitHub's default security settings prevent commits made by a bot's GITHUB_TOKEN from triggering subsequent workflows (to prevent infinite Action loops).If you want I can pull the bot's changes, fix the ruff errors locally, and push a new commit. Since this push came from my user account, the CI pipelines should all be re-triggered and visible now.

BitForge95:
@pollytur @reneburghardt Ruff was flagging B905 because I removed strict=True from the zip() functions. I had to remove strict=True to keep the tests compatible with Python 3.9 (which threw a TypeError in pytest). I added a # noqa: B905 to bypass the linter for those two lines so the CI can pass, but please let me know if you'd prefer i handle that 3.9 compatibility differently,

reneburghardt:
@BitForge95 Can you run tests with 3.12.8 (as in https://github.com/sensorium-competition/experanto/blob/d85148452d0af4df434c77dd814fdef70ac0e018/.github/workflows/test.yml#L31)? This should then allow to solve ruff issues, right?
Lets not worry about backcompatibility to 3.9 here (IMO); we might get rid of 3.9 anyway soon and if not, it needs to be handled in a different PR I think

Another thing that also came to my mind: we basically now have to two functions in create_experiment.py for returning config (`get_default_config` and `make_modality_config`) and `make_sequence_device` is also similar to `_generate_sequence_data` from `create_sequence_data.py`. Can we get rid of one config function and maybe `make_modality_config`? Otherwise we have similar functions that we all have to maintain in the future. I know this redundancy happened due to another PR, but would be nice if we can solve it here.

BitForge95:
Thanks @reneburghardt. I'll bump the CI to 3.12.8. I'll refactor the tests right now to use setup_test_experiment (with _generate_sequence_data) and get_default_config, and remove the redundant functions. I'll push those updates shortly.

reneburghardt:
> Thanks @reneburghardt. I'll bump the CI to 3.12.8. I'll refactor the tests right now to use setup_test_experiment (with _generate_sequence_data) and get_default_config, and remove the redundant functions. I'll push those updates shortly.

The CI is already at 3.12.8, but you mentioned 3.9 for tests, hence I assumed you ran tests locally with 3.9 leading to the issues you described

BitForge95:
Sorry , I meant locally . my bad

BitForge95:
@reneburghardt I just pushed the refactored tests using setup_test_experiment and get_default_config, and completely removed the redundant helper functions. Let me know if i need to do any more changes.

binary69:
Hey @reneburghardt just for the clarification, for `test_experiment_start_end_time_reflects_union` which is parametrized over 1, 2, and 3+ devices, `get_default_config` is hardcoded for exactly 2 devices named `device_0` and `device_1`. How does the refactored version handle the 1-device and 3-device cases without `make_modality_config`

binary69:
I noticed `pytest.approx` is used for the `start_time/end_time` comparisons here, is that intentional? I ask because in #137 the feedback was to use exact equality since the values come directly from `meta.yml` with no numerics involved. Happy to learn if there's a reason to prefer approx here:)

BitForge95:
Hi @binary69 , In the refactored test_experiment_start_end_time_reflects_union, I actually stopped using get_default_config for that exact reason. Instead, the test builds the config dynamically based on however many devices are passed in, so it scales perfectly to 1, 3, or N devices.

```config = {}
        for i in range(len(device_ranges)):
            config[f"device_{i}"] = {
                "interpolation": {
                    "sampling_rate": 10.0,
                    "offset": float(np.random.rand()),
                }
            }

        experiment = Experiment(
            root_folder=str(experiment_path), modality_config=config
        )
```
By looking at the #137 , you are right . I originally used it out of habit to stay safe against floating-point drift. Exact equality is preferred in this case. Thanks for linking #137 for context

reneburghardt:
> @BitForge95 Can you run tests with 3.12.8 (as in
> 
> https://github.com/sensorium-competition/experanto/blob/d85148452d0af4df434c77dd814fdef70ac0e018/.github/workflows/test.yml#L31
> 
> )? This should then allow to solve ruff issues, right?
> Lets not worry about backcompatibility to 3.9 here (IMO); we might get rid of 3.9 anyway soon and if not, it needs to be handled in a different PR I think
> Another thing that also came to my mind: we basically now have to two functions in create_experiment.py for returning config (`get_default_config` and `make_modality_config`) and `make_sequence_device` is also similar to `_generate_sequence_data` from `create_sequence_data.py`. Can we get rid of one config function and maybe `make_modality_config`? Otherwise we have similar functions that we all have to maintain in the future. I know this redundancy happened due to another PR, but would be nice if we can solve it here.

Sorry I just saw that I confused `make_modality_config` with `make_sequence_device` here. Originally, I wanted to keep either one of the config functions (so either `get_default_config` or `make_modality_config`, whatever makes more sense) and remove `make_sequence_device` if possible and use `_generate_sequence_data`. 

So if I see it correctly after a quick glance, due to my not well-written comment we now have `get_default_config` and another mechanism without a function to loop over multiple devices, right? As this are then still two different places of similar functionality which need to be maintained, I'd say we use `make_modality_config` instead of `get_default_config` (and extend it if necessary). What do you think @BitForge95 ? Sorry for not checking my earlier comment properly

BitForge95:
@reneburghardt , no worries. Now this makes this clear. make_modality_config is definitely the cleaner choice since it dynamically scales to any number of devices. I've just pushed a commit that removes get_default_config and refactors the tests to rely purely on make_modality_config and setup_test_experiment. Also according to the #137 I also removed pytest.approx in favor of exact equality == across the board in this latest push. Let me know how this looks ?

tests/create_sequence_data.py
```diff
@@ -1,83 +1,115 @@
...
 @contextmanager
 def sequence_data_and_interpolator(data_kwargs=None, interp_kwargs=None):
     data_kwargs = data_kwargs or {}
     interp_kwargs = interp_kwargs or {}
     with create_sequence_data(**data_kwargs) as (timestamps, data, shifts):
-        with closing(
-            Interpolator.create("tests/sequence_data", **interp_kwargs)
-        ) as seq_interp:
+        # Restore the helper expected by the rest of the test suite
+        from experanto.interpolators import Interpolator
```
reneburghardt: Please only do imports at the top of the file

tests/create_sequence_data.py
```diff
@@ -1,83 +1,115 @@
...
+
+        seq_interp = Interpolator.create(str(SEQUENCE_ROOT), **interp_kwargs)
+        try:
             yield timestamps, data, shifts, seq_interp
+        finally:
+            seq_interp.close()
```
reneburghardt: What is the reason to replace `closing` with the verbose code? I thought it should do the same, but closing is more compact

reneburghardt:
> @reneburghardt , no worries. Now this makes this clear. make_modality_config is definitely the cleaner choice since it dynamically scales to any number of devices. I've just pushed a commit that removes get_default_config and refactors the tests to rely purely on make_modality_config and setup_test_experiment. Also according to the #137 I also removed pytest.approx in favor of exact equality == across the board in this latest push. Let me know how this looks ?

Thanks! I had two remarks, otherwise so far I think it looks good

tests/create_sequence_data.py
BitForge95: Sorry , I actually used the try/finally block while i was debugging locally and I completely forgot about it. Yeah closing does the exact same thing , i will use closing.

BitForge95:
@reneburghardt Thank you for the review, I have made the changes , please let me know if I need to make any more changes.

reneburghardt (approved):
LGTM

pollytur:
@BitForge95 could we please fix ruff formatting?
 
https://github.com/BitForge95/experanto/actions/runs/24285020830/job/70912845714

image

tests/create_experiment.py
```diff
@@ -1,48 +1,30 @@
...
 
-import numpy as np
-import yaml
+from .create_sequence_data import _generate_sequence_data
 
 
 @contextmanager
-def make_sequence_device(
-    root, name, start, end, sampling_rate=10.0, n_signals=5, override_meta=None
+def setup_test_experiment(
```
pollytur, line 8: technically experiment could consist of not only sequence-like modalities hence this function needs to be extended - we will take it to a separate issue, so fine for now for the sake of this PR.

tests/test_experiment.py
```diff
@@ -32,19 +35,16 @@
...
 INVALID_META_CASES = [
-    {"start_time": None, "end_time": None},  # Both missing
-    {"start_time": None, "end_time": 10.0},  # Missing start_time
-    {"start_time": 0.0, "end_time": None},  # Missing end_time
-    {"start_time": float("inf"), "end_time": 10.0},  # Infinite start_time
-    {"start_time": 0.0, "end_time": float("inf")},  # Infinite end_time
-    {"start_time": float("-inf"), "end_time": 10.0},  # Negative Infinite start_time
-    {"start_time": 0.0, "end_time": float("-inf")},  # Negative Infinite end_time
-    {"start_time": float("nan"), "end_time": 10.0},  # NaN start_time
-    {"start_time": 0.0, "end_time": float("nan")},  # NaN end_time
```
pollytur, line 47: @BitForge95 why all these comments got removed? I would argue they were helpful for readability and suggest to plug them back

Did anybody suggested otherwise?

tests/create_sequence_data.py
```diff
@@ -10,74 +10,106 @@
...
-            fp.flush()  # Ensure data is written to disk
-            del fp
-        meta["dtype"] = str(data.dtype)
+    with open(sequence_root / "meta.yml", "w") as f:
+        yaml.safe_dump(meta, f)
 
-        if shifts_per_signal:
-            shifts = np.random.rand(n_signals) / meta["sampling_rate"] * 0.9
-            np.save(SEQUENCE_ROOT / "meta" / "phase_shifts.npy", shifts)
+    return timestamps, data, shifts
```
pollytur, line 79: @BitForge95 does it actually need to return this? I am not 1005 sure but looks like saving the data is enough and the direct return values are never used (and I guess they should not be used)

tests/test_experiment.py
pollytur: comments from line 20 till here

tests/test_experiment.py
```diff
@@ -70,137 +235,92 @@
...
-    ), f"Expected start_time={start_val}, got {experiment.start_time}"
-    assert experiment.end_time == (
-        end_val
-    ), f"Expected end_time={end_val}, got {experiment.end_time}"
-    assert any(
-        "invalid_device" in message for message in caplog.messages
-    ), "Expected warning about invalid_device was skipped"
+    # Removed pytest.approx here
+    assert experiment.start_time == start_val
+    assert experiment.end_time == end_val
```
pollytur: there was no approx in lines above, these are exactly same asserts as before but without the elaborate error messages, please revert it back
```suggestion
   
    assert experiment.start_time == (
        start_val
    ), f"Expected start_time={start_val}, got {experiment.start_time}"
    assert experiment.end_time == (
        end_val
    ), f"Expected end_time={end_val}, got {experiment.end_time}"
    assert any(
        "invalid_device" in message for message in caplog.messages
    ), "Expected warning about invalid_device was skipped"
```

tests/test_experiment.py
```diff
@@ -60,7 +60,172 @@
...
+    (tmp_path / "ignored_device").mkdir()
+    config = {
+        "screen": {"interpolation": mock_interpolator},
+        "eye_tracker": {"interpolation": mock_interpolator},
+    }
+    exp = Experiment(root_folder=str(tmp_path), modality_config=config)
+    assert "screen" in exp.devices
+    assert "eye_tracker" in exp.devices
+    assert "ignored_device" not in exp.devices
+    assert set(exp.device_names) == {"screen", "eye_tracker"}
```
pollytur: please add the error messages to the asserts (eg explanations if they fail), see how it was done before

tests/test_experiment.py
```diff
@@ -60,7 +60,172 @@
...
+    config = {"screen": {"interpolation": mock_interpolator}}
+    exp = Experiment(root_folder=str(tmp_path), modality_config=config)
+    test_times = np.array([10.0, 20.0])
+    res = exp.interpolate(test_times, device="screen")
+    mock_interpolator.interpolate.assert_called_once_with(
+        test_times, return_valid=False
+    )
+    np.testing.assert_array_equal(res, np.array([1, 2, 3]))
+    res_dict = exp.interpolate(test_times, device=None)
+    assert isinstance(res_dict, dict)
```
pollytur: please add the error messages to the asserts

tests/test_experiment.py
```diff
@@ -60,7 +60,172 @@
...
+        config["device_2"] = {
+            "sampling_rate": 1.0,
+            "chunk_size": 40,
+            "interpolation": {"interpolation_mode": "nearest_neighbor"},
+        }
+        experiment = Experiment(
+            root_folder=str(experiment_path), modality_config=config
+        )
+        valid_range = experiment.get_valid_range(device_name)
+        assert valid_range == (start_t, end_t)
```
pollytur: again assert error message

tests/test_experiment.py
```diff
@@ -70,137 +235,92 @@
...
 
-    assert experiment.start_time == (
-        expected_start
-    ), f"Expected start_time={expected_start}, got {experiment.start_time}"
-    assert experiment.end_time == (
-        expected_end
-    ), f"Expected end_time={expected_end}, got {experiment.end_time}"
+    # Removed pytest.approx here
+    assert experiment.start_time == expected_start
+    assert experiment.end_time == expected_end
```
pollytur: there was no pytest.approx above, these are same assertions but without error messages, please revert back!

tests/test_experiment.py
```diff
@@ -70,137 +235,92 @@
...
 
         with caplog.at_level(logging.WARNING, logger="experanto.experiment"):
             experiment = Experiment(
-                root_folder=tmp_path,
-                modality_config=make_modality_config("valid_device", "invalid_device"),
+                root_folder=str(experiment_path), modality_config=config
             )
 
     assert "valid_device" in experiment.devices
     assert "invalid_device" not in experiment.devices
```
pollytur: please add error messages in case asserts fail

tests/test_experiment.py
```diff
@@ -70,137 +235,92 @@
...
-    log a warning, but still initialize successfully if at least one valid
-    device is present. The experiment time range should reflect only the
-    valid device.
-    """
-    start_val = np.random.lognormal(mean=0.0, sigma=1.0)  # Strictly positive float
-    duration_val = np.random.lognormal(mean=0.0, sigma=1.0)
+    start_val, duration_val = (
+        np.random.lognormal(0.0, 1.0),
+        np.random.lognormal(0.0, 1.0),
+    )
```
pollytur: why do we have this changes here? again, comments were helpful imho and the other change is not meaningful - please avoid changing sth without a good reason as it makes finding meaningful changes really hard through the git history
revert it please

tests/test_experiment.py
```diff
@@ -70,137 +235,92 @@
...
-                end=end_nonval,
-                override_meta=override_meta,
-            )
-        )
+    devices_kwargs = [
+        {"start_time": start_val, "t_end": end_val},  # valid device
+        {"start_time": 0.0, "t_end": 10.0},  # invalid device
+    ]
+
+    with setup_test_experiment(
```
pollytur, line 349: i might be missing sth but why it got changed from `with ExitStack() as stack: ...` to this?

tests/test_experiment.py
```diff
@@ -70,137 +235,92 @@
...
+        {
+            "start_time": start,
+            "t_end": end,
+            "n_signals": n_signals,
+            "sampling_rate": float(np.random.randint(5, 30)),
+        }
+        for start, end in device_ranges
+    ]
+
+    with setup_test_experiment(
```
pollytur, line 286: i might be missing sth but why it got changed from with ExitStack() as stack: ... to this?

pollytur (changes requested):
@BitForge95 please see the comments above :)
And thanks a lot for iterating on it!

tests/create_experiment.py
BitForge95: I’ll leave this as it is for this PR, but yeah, tracking that extension in a separate issue makes sense.

tests/test_experiment.py
BitForge95: Looks like my auto-formatter got a bit too aggressive during the refactor and removed these by mistake. I’ve put those lines back so the comments and formatting are exactly how they were before.

tests/create_sequence_data.py
BitForge95: Yess! Since the data is being saved directly to disk, there’s no need to return anything. I’ve removed the return statement to keep things simple.

tests/test_experiment.py
BitForge95: Sorry, my formatter changed this block unintentionally. I have reverted it back to the exact original state in the latest commit.

tests/test_experiment.py
BitForge95: The new setup_test_experiment context manager natively handles spinning up multiple devices at once via the n_devices parameter. This allowed us to drop the ExitStack loops and make the test setups much more compact.

BitForge95:
@pollytur, the CI just failed on the Run Tests step, specifically on test_linear_interpolation.

Looking at the logs, it seems to be a statistical edge case: because contain_nans=True, the generator occasionally produces a column that is 100% NaNs. Since np.nanmean can't impute an entirely empty column, the final isnan().sum() == 0 assertion fails.

Would you like me to push a quick update to the test to exclude fully NaN columns from that final check so the CI passes, or would you prefer to handle that edge case differently?

pollytur:
@BitForge95 could you please add some logs screenshot for the nans behaviour that you describe above? 
and the pseudocode for the suggested fix

BitForge95:
image

Here is the screenshot of the CI logs showing the failure.

Because contain_nans=True, the generator happened to create a sequence where the second signal (column) was entirely NaNs. You can see in the boolean array output at the bottom that the second column is True all the way down.

Since np.nanmean cannot impute a column that is 100% NaN, those NaNs correctly remain in the interp array, which causes the final blanket sum() == 0 assertion to fail.

The Suggested Fix:
We just need to calculate which columns were 100% NaN in the expected data, and exclude those specific columns from the final assertion.

``` 
    if not keep_nans:
            # --- Current ---
            # assert np.isnan(interp).sum() == 0
            
            # --- Suggested Fix ---
            # Identify columns that are entirely NaN and cannot be imputed
            all_nan_cols = np.all(np.isnan(expected), axis=0)
            
            # Assert 0 NaNs only in the columns that were capable of being imputed
            assert (
                np.isnan(interp[:, ~all_nan_cols]).sum() == 0
            ), "Imputable signals should not contain NaNs"
```

tests/create_sequence_data.py
pollytur: I dont see it removed. Have you committed and pushed this changes?

tests/test_experiment.py
pollytur: @BitForge95

tests/test_experiment.py
```diff
@@ -161,40 +321,41 @@ def test_experiment_skips_invalid_devices(tmp_path, override_meta, caplog):
...
-                tmp_path,
-                "invalid_device",
-                start=start_nonval,
-                end=end_nonval,
-                override_meta=override_meta,
-            )
-        )
+    devices_kwargs = [
+        {"start_time": start_val, "t_end": end_val},  # valid device
+        {"start_time": 0.0, "t_end": 10.0},  # invalid device
```
pollytur: why is this 0, 10 are now hardcoded here?

tests/test_experiment.py
```diff
@@ -161,40 +321,41 @@ def test_experiment_skips_invalid_devices(tmp_path, override_meta, caplog):
...
-    assert "valid_device" in experiment.devices
-    assert "invalid_device" not in experiment.devices
+    assert (
+        "valid_device" in experiment.devices
+    ), "Expected 'valid_device' to be initialized"
+    assert (
+        "invalid_device" not in experiment.devices
+    ), "Expected 'invalid_device' to be skipped"
 
+    # Removed pytest.approx here
```
pollytur:
```suggestion
```

tests/test_experiment.py
```diff
@@ -161,40 +321,41 @@ def test_experiment_skips_invalid_devices(tmp_path, override_meta, caplog):
...
             )
 
-    assert "valid_device" in experiment.devices
-    assert "invalid_device" not in experiment.devices
+    assert (
+        "valid_device" in experiment.devices
+    ), "Expected 'valid_device' to be initialized"
+    assert (
+        "invalid_device" not in experiment.devices
+    ), "Expected 'invalid_device' to be skipped"
```
pollytur: @BitForge95 
Please redo all the other asserts in the same way, that the user could see the variables as it often helps to see what went wrong
```suggestion
    assert (
        "valid_device" in experiment.devices
    ), f"Expected 'valid_device' to be in experiment.devices but experiment.devices = {experiment.devices}"
    assert (
        "invalid_device" not in experiment.devices
    ), "Expected 'invalid_device' to be skipped in experiment.devices, instead experiment.devices = {experiment.devices}"
```

tests/create_sequence_data.py
BitForge95: Sorry, I mentioned it in the comments but completely forgot to actually stage that specific file in my last commit. Just pushed the removal now.

tests/test_experiment.py
BitForge95: We swapped out ExitStack because the new setup_test_experiment context manager natively handles spinning up multiple devices simultaneously via the n_devices parameter. It allowed us to drop the ExitStack loops entirely and make the test setups much more compact.

tests/test_experiment.py
```diff
@@ -70,32 +256,30 @@
...
 
         experiment = Experiment(
-            root_folder=tmp_path,
-            modality_config=make_modality_config(
-                *device_names, offsets=[float(np.random.rand()) for _ in device_names]
-            ),
+            root_folder=str(experiment_path), modality_config=config
         )
 
+    # Removed pytest.approx here
```
pollytur:
```suggestion
```

tests/test_experiment.py
```diff
@@ -60,7 +63,190 @@
...
+        ],
+    ) as experiment_path:
+        experiment = Experiment(
+            root_folder=str(experiment_path),
+            modality_config=make_modality_config("device_0"),
+        )
+        valid_range = experiment.get_valid_range("device_0")
+        # Keep pytest.approx here as it's dealing with floats generated by Hypothesis
+        assert valid_range[0] == pytest.approx(start_offset)
+        assert valid_range[1] == pytest.approx(start_offset + duration)
```
pollytur: please add error messages for the asserts

tests/test_experiment.py
```diff
@@ -60,7 +63,190 @@
...
+    config = {
+        "screen": {"interpolation": mock_interpolator},
+        "eye_tracker": {"interpolation": mock_interpolator},
+    }
+    exp = Experiment(root_folder=str(tmp_path), modality_config=config)
+    assert "screen" in exp.devices, "Expected 'screen' in experiment devices"
+    assert "eye_tracker" in exp.devices, "Expected 'eye_tracker' in experiment devices"
+    assert (
+        "ignored_device" not in exp.devices
+    ), "Expected 'ignored_device' to be excluded"
```
pollytur:
```suggestion
    assert "screen" in exp.devices, f"Expected 'screen' in experiment devices but exp.devices = {exp.devices}"
    assert "eye_tracker" in exp.devices, f"Expected 'eye_tracker' in experiment devices but exp.devices = {exp.devices}"
    assert (
        "ignored_device" not in exp.devices
    ), f"Expected 'ignored_device' to be excluded from experiment devices but exp.devices = {exp.devices}"
```

tests/test_experiment.py
BitForge95: They were originally just acting as dummy placeholders for the invalid device, since its metadata gets explicitly corrupted and overwritten by the INVALID_META_CASES anyway. However, to keep the test consistent, I'll update the file to just dynamically generate random values for it exactly like we do for the valid device.

tests/test_experiment.py
```diff
@@ -1,25 +1,28 @@
...
     ([(1.0, 8.0), (0.0, 10.0)], 0.0, 10.0),
-    # Three devices with different ranges: start should be min, end should be max
     ([(0.0, 10.0), (1.0, 8.0), (2.0, 9.0)], 0.0, 10.0),
-    # Devices with non-overlapping ranges: start should be min, end should be max
     ([(0.0, 3.0), (7.0, 8.0)], 0.0, 8.0),
-    # Devices with identical ranges: start and end should match that range
     ([(1.0, 5.0), (1.0, 5.0)], 1.0, 5.0),
-    # Large time stamps: start should be min, end should be max
     ([(1e9, 1e9 + 100), (1e9 - 50, 1e9 + 50)], 1e9 - 50, 1e9 + 100),
 ]
```
pollytur, line 33: please plug these comments back (revert the changes)

tests/test_experiment.py
```diff
@@ -161,46 +330,52 @@ def test_experiment_skips_invalid_devices(tmp_path, override_meta, caplog):
     duration_val = np.random.lognormal(mean=0.0, sigma=1.0)
     end_val = start_val + duration_val
 
-    start_nonval = np.random.lognormal(mean=0.0, sigma=1.0)
-    duration_nonval = np.random.lognormal(mean=0.0, sigma=1.0)
-    end_nonval = start_nonval + duration_nonval
```
pollytur, line 166: please revert this, as then the naming is consistent and in the edit below 10 is still hardcoded

tests/test_experiment.py
```diff
@@ -161,46 +330,52 @@ def test_experiment_skips_invalid_devices(tmp_path, override_meta, caplog):
...
-            )
-        )
+    # Generate random values for the invalid device as well
+    inv_start = np.random.lognormal(mean=0.0, sigma=1.0)
+    inv_end = inv_start + 10.0
+
+    devices_kwargs = [
+        {"start_time": start_val, "t_end": end_val},  # valid device
+        {"start_time": inv_start, "t_end": inv_end},  # invalid device
+    ]
```
pollytur:
```suggestion
    start_nonval = np.random.lognormal(mean=0.0, sigma=1.0)
    duration_nonval = np.random.lognormal(mean=0.0, sigma=1.0)
    end_nonval = start_nonval + duration_nonval

    devices_kwargs = [
        {"start_time": start_val, "t_end": end_val},  # valid device
        {"start_time": start_nonval, "t_end": end_nonval},  # invalid device
    ]
```

tests/test_experiment.py
pollytur: @reneburghardt do you have an opinion on it?

pollytur:
@BitForge95 seems like the return in `create_sequence_data.py‎` was needed cause now a lot of tests fail https://github.com/BitForge95/experanto/actions/runs/24388879059/job/71229619076

image

tests/create_sequence_data.py
```diff
@@ -10,74 +10,104 @@
...
+    data = np.random.rand(len(timestamps), n_signals)
 
-        if shifts_per_signal:
-            shifts = np.random.rand(n_signals) / meta["sampling_rate"] * 0.9
-            np.save(SEQUENCE_ROOT / "meta" / "phase_shifts.npy", shifts)
+    if contain_nans:
+        nan_indices = np.random.choice(
+            data.size, size=int(0.1 * data.size), replace=False
+        )
+        data.flat[nan_indices] = np.nan
```
pollytur, line 51: @BitForge95 thanks a lot for the failing case of the test that you found! We will make an issue and discuss with the team how we want to handle the cases when effectively there is no signal (all values over time are Nans) - for now lets instead fix the data generation to have at least 1 non-Nan value per row


```suggestion
        )
        data.flat[nan_indices] = np.nan
        
        # ensure each row has at least one non-NaN
        row_all_nan = np.isnan(data).all(axis=1)
        data[row_all_nan, 0] = 0.0
```

tests/test_experiment.py
BitForge95: I reverted the naming back to start_nonval / duration_nonval and completely removed that leftover hardcoded 10.0. It uses the np.random generators now to match the valid device setup.

tests/create_sequence_data.py
pollytur: @BitForge95 i think thats the last thing left from my side

Thanks a lot for your fast iterations and your patience with me being pedantic!

tests/create_experiment.py
pollytur: I created issue #159 hence resolve the thread here

tests/create_sequence_data.py
```diff
@@ -50,6 +50,10 @@ def _generate_sequence_data(
         )
         data.flat[nan_indices] = np.nan
 
+    # ensure each row has at least one non-NaN
+    row_all_nan = np.isnan(data).all(axis=1)
+    data[row_all_nan, 0] = 0.0
+
```
pollytur: @BitForge95 the tabs are wrong here...
It should be under 
`if contain_nans:`

BitForge95: Ohh yeah sorry , i guess that is why the tests are failing. Thank you for the catch !

pollytur (approved):
@BitForge95 thanks a lot once again!

PR #126, opened 2026-03-10
Add neuron selection support to SequenceInterpolator and SpikeInterpolator

Vrittigyl:
## Description

This PR adds support for selecting specific neurons in both `SequenceInterpolator` and `SpikeInterpolator`.

Users can now pass either:
- `neuron_ids`: biological neuron IDs mapped using `meta/unit_ids.npy`
- `indexes`: direct neuron indexes

If both are provided:
- A warning is raised if they refer to the same neurons.
- A `ValueError` is raised if they refer to different neurons.

### Changes
- **SequenceInterpolator**
  - Added neuron filtering using `neuron_ids` or `indexes`.
  - Filters loaded data and normalization statistics to match selected neurons.

- **PhaseShiftedSequenceInterpolator**
  - Filters `phase_shifts` to match selected neurons when neuron selection is used.

- **SpikeInterpolator**
  - Added neuron filtering using `neuron_ids` or `indexes`.
  - Rebuilds the spike array and indices to include only selected neurons.

This allows loading and processing only a subset of neurons, reducing memory usage and improving flexibility when working with imaging datasets.

Closes #124

### DIFF
--- a/experanto/interpolators.py
+++ b/experanto/interpolators.py
@@ -20,6 +20,52 @@ from .intervals import TimeInterval
 logger = logging.getLogger(__name__)
 
 
+def resolve_neuron_indices(neuron_ids, neuron_indices, unit_ids, n_signals):
+    if neuron_ids is None and neuron_indices is None:
+        return None
+
+    if neuron_ids is not None:
+        ids_to_indexes = []
+
+        for nid in neuron_ids:
+            match = np.where(unit_ids == nid)[0]
+            if len(match) == 0:
+                raise ValueError(f"Neuron id {nid} not found")
+            ids_to_indexes.append(int(match[0]))
+
+        if neuron_indices is None:
+            return ids_to_indexes
+
+        if set(ids_to_indexes) != set(neuron_indices):
+            raise ValueError("neuron_ids and neuron_indices refer to different neurons")
+
+        warnings.warn(
+            "Both neuron_ids and neuron_indices provided; using neuron_indices",
+            stacklevel=2,
+        )
+
+    return validate_neuron_indices(neuron_indices, n_signals)
+
+
+def validate_neuron_indices(neuron_indices, n_signals):
+    try:
+        indexes_seq = list(neuron_indices)
+    except TypeError as exc:
+        raise TypeError("neuron_indices must be iterable") from exc
+
+    if not all(isinstance(i, (int, np.integer)) for i in indexes_seq):
+        raise TypeError("neuron_indices must contain integers")
+
+    if indexes_seq:
+        if min(indexes_seq) < 0 or max(indexes_seq) >= n_signals:
+            raise ValueError("neuron_indices out of bounds")
+
+        if len(set(indexes_seq)) != len(indexes_seq):
+            raise ValueError("neuron_indices contain duplicates")
+
+    return indexes_seq
+
+
 class Interpolator:
     """Abstract base class for time series interpolation.
 
@@ -55,8 +101,10 @@ class Interpolator:
     Experiment : High-level interface that manages multiple interpolators.
     """
 
-    def __init__(self, root_folder: str) -> None:
+    def __init__(self, root_folder: str | Path) -> None:
         self.root_folder = Path(root_folder)
+        self.n_signals: int = 0
+        self.interpolation_mode: str | None = None
         self.start_time = None
         self.end_time = None
         # Valid interval can be different to start time and end time.
@@ -177,6 +225,14 @@ class SequenceInterpolator(Interpolator):
         If True, subtracts mean during normalization.
     normalize_std_threshold : float, optional
         Minimum std threshold to prevent division by near-zero values.
+    neuron_ids : list, optional
+        Biological neuron IDs to include. Converted to indexes using meta/unit_ids.npy.
+    neuron_indices : list, optional
+        Column indexes of neurons to include.
+    neuron_ids : list, optional
+        Biological neuron IDs to include. Converted to indexes using meta/unit_ids.npy.
+    neuron_indices : list, optional
+        Column indexes of neurons to include.
     **kwargs
         Additional keyword arguments (ignored).
 
@@ -202,10 +258,12 @@ class SequenceInterpolator(Interpolator):
 
     def __init__(
         self,
-        root_folder: str,
+        root_folder: str | Path,
         cache_data: bool = False,  # already cached, put it here for consistency
         keep_nans: bool = False,
         interpolation_mode: str = "nearest_neighbor",
+        neuron_ids: list[int] | None = None,
+        neuron_indices: list[int] | None = None,
         normalize: bool = False,
         normalize_subtract_mean: bool = False,
         normalize_std_threshold: float | None = None,  # or 0.01
@@ -215,6 +273,7 @@ class SequenceInterpolator(Interpolator):
         meta = self.load_meta()
         self.keep_nans = keep_nans
         self.interpolation_mode = interpolation_mode
+        self.neuron_ids = neuron_ids
         self.normalize = normalize
         self.normalize_subtract_mean = normalize_subtract_mean
         self.normalize_std_threshold = normalize_std_threshold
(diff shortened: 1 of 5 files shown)


experanto/interpolators.py
```diff
@@ -223,28 +281,41 @@ def __init__(
...
             )
-
-            if cache_data:
-                self._data = np.array(self._data).astype(
-                    np.float32
-                )  # Convert memmap to ndarray
         else:
             self._data = np.load(self.root_folder / "data.npy")
 
+        # Apply indexing BEFORE caching
```
reneburghardt: Hi @pollytur 
as far as I know, if one applies indexes to memmap as lists or arrays, this loads a copy into RAM (see https://stackoverflow.com/questions/18614927/how-to-slice-memmap-efficiently or https://stackoverflow.com/questions/78426050/how-to-index-a-numpy-memmap-without-creating-an-in-memory-copy)
Only regular continuous slicing creates just a view, but I don't think neuron ids will always be continuous.

Have you thought about this? We might need to set caching to True if neurons are indexed. Or we find a workaround (I haven't investigated it yet).

pollytur: great catch, no I have not thought about it tbh
we probably want to investigate the workaround (changing order of neurons as we need and save it as a temp memmap file is the first though but thats insanly memory inefficient since neuronal responses are also the heaviest part of the dataset from a memory perspective...)

Vrittigyl: Or maybe instead of directly indexing, we could iteratively fetch only the required columns (or in chunks) and optionally cache them.

reneburghardt: @Vrittigyl could you please provide an implementation for what you proposed and show that it still caches the data if cache_data == True and doesn't cache the data otherwise, also when we select only certain neurons? Thanks!

experanto/interpolators.py
```diff
@@ -67,6 +69,53 @@ def load_meta(self):
...
+            raise TypeError("neuron_indices must contain integers")
+
+        if indexes_seq:
+            if min(indexes_seq) < 0 or max(indexes_seq) >= self.n_signals:
+                raise ValueError("neuron_indices out of bounds")
+
+            if len(set(indexes_seq)) != len(indexes_seq):
+                raise ValueError("neuron_indices contain duplicates")
+
+        return indexes_seq
```
reneburghardt: Sorry, but I don't like putting those functions into Interpolator abstract class as these functions only make sense for neuron-related interpolators. Can we move these functions outside of any class and just pass all needed params into the function (as we can't use `self` then anymore)? I would then load `unit_ids` before calling `_resolve_indices` and pass `unit_ids` directly instead of `self.root_folder`.

Vrittigyl: Will surely fix it sir

experanto/interpolators.py
Vrittigyl: I’ve implemented this and tested it using test functions.
When `cache_data=False`, `_data` stays as memmap and only required neuron columns are accessed during interpolation. When `cache_data=True`, only selected neurons are loaded into memory.
I verified this by checking data types, shapes, and comparing outputs.

pollytur:
@Vrittigyl could you please take a look why tests are failing now?

Vrittigyl:
> @Vrittigyl could you please take a look why tests are failing now?

 _data was expected to be sliced after neuron selection, but now _data stays full to support memmap.
I’ve updated the tests to check n_signals and interpolation output instead.

pollytur:
@Vrittigyl please merge current main into your branch - I resolved the conflict but its not a full merge, so ruff CI / CD fails because of it

please ping me after its done and I will take a look

Vrittigyl:
> @Vrittigyl please merge current main into your branch - I resolved the conflict but its not a full merge, so ruff CI / CD fails because of it
> 
> please ping me after its done and I will take a look

@pollytur I’ve merged the latest main into my branch, resolved the conflicts, and fixed the CI issues. Please take a look!

pollytur:
@Vrittigyl sorry for back and forward
could we please do it again (merge main into your branch) since I have just merged #113 in main and that PR added a bunch of tests which i would love to see here as well

Also you have been editing same files so merging is quite crucial here

tests/create_screen_data.py
```diff
@@ -50,7 +50,7 @@ def create_screen_data(
                 chunk = video_frames[start:end]
                 if len(chunk) == 0:
                     continue
-                video_array = np.stack(chunk, axis=0)
+                video_array = np.stack(list(chunk), axis=0)
```
pollytur, line 53: why do we need it?

tests/test_sequence_interpolator.py
```diff
@@ -130,6 +130,7 @@ def test_nearest_neighbor_interpolation_with_phase_shifts(
             "shifts_per_signal": True,
         }
     ) as (timestamps, data, shift, seq_interp):
+        assert shift is not None
```
pollytur: please add error message in case the assert fails, ideally with f -string and values if needed

tests/test_sequence_interpolator.py
```diff
@@ -282,6 +283,7 @@ def test_linear_interpolation_with_phase_shifts(
         },
         interp_kwargs={"keep_nans": keep_nans},
     ) as (timestamps, data, shift, seq_interp):
+        assert shift is not None
```
pollytur: please add error message in case the assert fails, ideally with f -string and values if needed

tests/test_sequence_interpolator.py
```diff
@@ -389,6 +391,7 @@ def test_interpolation_with_phase_shifts_for_invalid_times(
         },
         interp_kwargs={"keep_nans": keep_nans},
     ) as (_, _, phase_shifts, seq_interp):
+        assert phase_shifts is not None
```
pollytur: please add error message in case the assert fails, ideally with f -string and values if needed

tests/test_spikes_interpolator.py
```diff
@@ -124,7 +124,7 @@ def test_spikes_cache_data():
         data_kwargs={"duration": 5.0, "n_neurons": 2},
         interp_kwargs={"cache_data": True},
     ) as (gt_spikes, interp):
-
+        assert isinstance(interp, SpikeInterpolator)
```
pollytur: please add error message in case the assert fails, ideally with f -string and values if needed

tests/test_spikes_interpolator.py
```diff
@@ -169,6 +169,7 @@ def test_memmap_loading():
         },
         interp_kwargs={"cache_data": False},
     ) as (gt_spikes, interp):
+        assert isinstance(interp, SpikeInterpolator)
```
pollytur: please add error message in case the assert fails, ideally with f -string and values if needed

tests/test_spikes_interpolator.py
```diff
@@ -186,10 +187,52 @@ def test_memmap_loading():
         },
         interp_kwargs={"cache_data": True},
     ) as (gt_spikes, interp):
+        assert isinstance(interp, SpikeInterpolator)
```
pollytur: please add error message in case the assert fails, ideally with f -string and values if needed

tests/test_time_intervals_interpolator.py
```diff
@@ -79,7 +79,8 @@ def run_interval_interpolation_test(timestamps, intervals_dict, time_interp):
     assert isinstance(time_interp, TimeIntervalInterpolator)
 
     signal = time_interp.interpolate(timestamps)
-
+    if isinstance(signal, tuple):
+        signal = signal[0]
```
pollytur, line 83: why do we need it here and everywhere else in this file?

experanto/interpolators.py
```diff
@@ -20,6 +20,52 @@
 logger = logging.getLogger(__name__)
 
 
+def resolve_neuron_indices(neuron_ids, neuron_indices, unit_ids, n_signals):
```
pollytur, line 23: imho `resolve_neuron_indices` should go to utils.py
@reneburghardt what would you say?

experanto/interpolators.py
```diff
@@ -20,6 +20,52 @@
...
+
+        warnings.warn(
+            "Both neuron_ids and neuron_indices provided; using neuron_indices",
+            stacklevel=2,
+        )
+
+    return validate_neuron_indices(neuron_indices, n_signals)
+
+
+def validate_neuron_indices(neuron_indices, n_signals):
```
pollytur, line 50: imho `validate_neuron_indices` should go to utils.py
@reneburghardt what would you say?

I am also tempted to put some function like `select_channels` to `SequenceInterpolator` class. Because in principle we might also want to select only certain channels from eye tracker as well. And a particular case of this function would call the function from utils for the neurons (to resolve unit_ids meta to indexes and validate it)

tests/create_screen_data.py
Vrittigyl: I changed this because the previous version was causing pyright failures with `np.stack` when chunk wasn’t a proper sequence. Converting it to `list(chunk)` ensures it works consistently.

experanto/interpolators.py
reneburghardt: Yes fine for me

experanto/interpolators.py
reneburghardt: Yes

Vrittigyl: Sorry for the late reply. I'll move `validate_neuron_indices` to utils.py
Should I also implement `select_channels` in `utils.py`, or keep it inside `SequenceInterpolator`?

Vrittigyl: Right now we use `neuron_indices`, but if we add something like `select_channels` for other modalities (like eye tracker), the idea is basically the same - selecting columns using indices.
So would it make sense to rename `neuron_indices` to `channel_indices`?
or is it better to keep `neuron_indices` separate?

Also, am I thinking about this the right way?

PR #147, opened 2026-03-19
Fix doc building and version constraints and allow manual dispatching of CI pipelines

reneburghardt:
## Description
- After introduction of spikes interpolator, the building of the docs failed because readthedocs is building with python v3.9 which doesn't support numba in the specified version
- The CI failed to detect this, because it runs in v3.12
- I also added black and isort to dev dependency group with version constraints
- constrained the version of pyright
- I make use of the dev dependency group in CI to have a single place of definition for dependency versions
- I renamed "experanto" string in readthedocs to capitalized version, matching it to other places
- I introduced manual dispatching of CI workflows
- The name and email of Github Actions Bot was adjusted according to https://github.com/actions/checkout?tab=readme-ov-file#push-a-commit-to-a-pr-using-the-built-in-token

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - '**'
+  workflow_dispatch:
 
 jobs:
   build-docs:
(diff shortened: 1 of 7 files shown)


pollytur (approved):
LGTM

oliveira-caio (approved):

PR #151, opened 2026-03-20
Add NWB to Experiment conversion using tiers modality

Vrittigyl:
This PR extends the existing DANDI example by demonstrating how to construct an Experiment object from NWB data.

The main addition is a pipeline that extracts TimeIntervals (e.g. trials or epochs) from NWB files and converts them into a "tiers" modality. These intervals are grouped based on a label column (either user-defined or automatically selected).

For automatic selection, the implementation currently chooses the column with the fewest unique values among valid candidates.

Key points:

- Supports different NWB structures (trials / epochs) without hardcoding
- Automatically selects a suitable label column when available (based on minimum unique values)
- Builds a tiers/ folder with interval data and metadata
- Uses TimeIntervalInterpolator to enable time-based querying
- Demonstrates usage through exp.interpolate(...)

This fills the missing step in the current DANDI example by showing how NWB data can be integrated into the Experiment .


pollytur (changes requested):
Thanks @Vrittigyl  for taking this in work!


Crucial:

**The experiment creating is not actually correct**

```
labels, t_start, t_end = extract_intervals_from_nwb(nwb128)

root = build_tiers_folder(labels, t_start, t_end)

config = dict(DEFAULT_MODALITY_CONFIG)
config["tiers"] = {
    "interpolation": {
        "_target_": "experanto.interpolators.TimeIntervalInterpolator"
    }
}

exp = Experiment(root_folder=root, modality_config=config)
print(exp.device_names) 

times = np.linspace(t_start, t_end, 10000)
result, valid = exp.interpolate(times, device="tiers", return_valid=True)

print(result.shape)       
print(result.sum(axis=0)) 
```
Here the only device / modality is the tiers, while in fact all the modalities from this experiment should be a part of the created experiment + the tiers modality.


Minor:

*. check if `!pip install -q "numba>=0.61.0" --upgrade` is still needed since `"numba==0.63.1"` is in the toml dependencies
* the pep guidelines are not there sometimes - like `def build_tiers_folder(` should have 2 empty lines before 
* import of `import h5py` should be at the 2nd cell and not in the middle

reneburghardt:
Just a quick thing I noticed: the Colab button is not updated, is it?

Vrittigyl:
I was planning to update the colab once the approach is approved.

pollytur:
@Vrittigyl please ping the reviewers next time and re-request changes if the notebook if the changes are ready for another review. Unfortunately, we cannot often open all of the ongoing PRs and check if its ready for a re-review or not

pollytur:
@Vrittigyl 

> Instead of manually defining each modality, detect_modalities could return a ready-to-use config (with interpolator + defaults). 

Do you mean thats instead of doing sth like in cell 12

```
config = dict(DEFAULT_MODALITY_CONFIG)
config["tiers"] = {
    "interpolation": {
        "_target_": "experanto.interpolators.TimeIntervalInterpolator"
    }
}
config["spikes"] = {
    "interpolation": {
        "_target_": "experanto.interpolators.SpikeInterpolator",
        "cache_data": True,
        "interpolation_window": 0.1,
        "interpolation_align": "center",
    }
}
```

to have this predefined in the `detect_modality` function? Maybe a small pseudocode would help me to understand you

And sorry for the long reply, please do not hesitate to ping us!

pollytur (changes requested):
@Vrittigyl thanks a lot for the nice notebook. I guess functionally its fine with me 

One cosmetics thing I would kindly ask to add :
at the beginning on the notebook in the markdown cell before the `pip install ...` cell could you please add the explanation that we also use `TimeIntervalInterpolator` for tiers and that a) its an optional modality but its is crucial for deep learning models that train-validation-test splits are fixed and the model results are comparable. If no DL model is needed tiers modality could be skips b) some datasets provide this split automatically like https://data.sensorium-competition.net/ (soon to be published) but most of dandi datasets dont have a fixed train-val-test split hence we create it here

If you think some other comments would be helpful - go for it

And change the colab link as well please

Vrittigyl:
> @Vrittigyl
> 
> > Instead of manually defining each modality, detect_modalities could return a ready-to-use config (with interpolator + defaults).
> 
> Do you mean thats instead of doing sth like in cell 12
> 
> ```
> config = dict(DEFAULT_MODALITY_CONFIG)
> config["tiers"] = {
>     "interpolation": {
>         "_target_": "experanto.interpolators.TimeIntervalInterpolator"
>     }
> }
> config["spikes"] = {
>     "interpolation": {
>         "_target_": "experanto.interpolators.SpikeInterpolator",
>         "cache_data": True,
>         "interpolation_window": 0.1,
>         "interpolation_align": "center",
>     }
> }
> ```
> 
> to have this predefined in the `detect_modality` function? Maybe a small pseudocode would help me to understand you
> 
> And sorry for the long reply, please do not hesitate to ping us!

Yes, exactly! the idea was to reduce repeating the same config blocks across datasets.

I wasn’t thinking of moving all config logic inside `detect_modalities`, but rather using it for detection and reusing the output to construct configs.
we can do something like:
```
#Add tiers config explicitly...
#then
modalities = detect_modalities(nwb)
for name, interp in modalities.items():
    config[name] = {"interpolation": {"_target_": f"experanto.interpolators.{interp}"}}
```
Also thanks for the detailed reply, it was really helpful! I will make sire to ping next time🙂

pollytur:
@Vrittigyl and what about 
```
"cache_data": True,
"interpolation_window": 0.1,
"interpolation_align": "center",
```

for the SpikeInterpolator?

Vrittigyl:
@pollytur For this I was thinking of either handling `SpikeInterpolator` as a special case and including these defaults, or keeping spikes explicitly defined and use auto-detection for the rest.

pollytur:
@Vrittigyl sure go for it and ping me when its done :)

just keep the defaults needed as they were

pollytur:
@Vrittigyl if you dont plan to update it - please let us know
and does the colab link now leads to lates up-to-date notebook version?

Vrittigyl:
Hey! I’ve updated the Colab notebook, it’s now the latest version.
Will add the proposed function soon as well. Got caught up with exams, sorry for the delay 😅 getting back to it now.

pollytur:
> Hey! I’ve updated the Colab notebook, it’s now the latest version. Will add the proposed function soon as well. Got caught up with exams, sorry for the delay 😅 getting back to it now.

no worries, I was just not sure if you still plan to update it :)
Good luck with the exams!

Vrittigyl:
@pollytur Thanks! I have pushed the changes now

pollytur:
@Vrittigyl is this the final version we should review? (as you have not re-requested the review)

pollytur (changes requested):
@Vrittigyl I caugh one conceptual problem with the notebook

Right now the interpolation is done via interpolator objects, which is redundant - everything should be done via the experiment object with `exp.interpolate` https://github.com/sensorium-competition/experanto/blob/main/experanto/experiment.py#L152 and not via re-creating an interpolator object outside of experiment, which is redundunt. This is the whole point of having an experiment class- to have all interpolators together

Another small thing - it would be nice to add a visualization of tiers, e.g. three lines one under each other which are 1 if this tier is true and 0 is this tier is false showing that train-val-test tiers are random chunks in time and non-overlapping

Vrittigyl:
@pollytur I’ve updated the notebook to use `exp.interpolate` directly instead of creating interpolators separately.
I had to map the modality names because the folder names didn’t match what `detect_modalities()` returns, so I’m not sure if that’s the right approach 😅

For the tiers visualization, what I’m currently getting is:
image

this is based on the existing labels (direction classes), not train/val/test. Should I create train/val/test tiers for this, or is there already some function that handles this split?

Vrittigyl:
Hi @pollytur , could you please take a look at this?

pollytur:
@Vrittigyl sorry for the long response, I was out-of-office
Could you please check why the docs building is failing now and fix it?

Vrittigyl:
@pollytur the docs build seems to have failed due to `SSL error`, unrelated to the notebook changes. Ig re-running the workflow should fix it.

PR #157, opened 2026-03-29
Add __all__ Export List to __init__.py 

spmridula:
### Summary
This PR improves the public API of Experanto by populating the previously empty `__init__.py` file. With these changes, users can now import core classes directly from the top-level package (e.g., `from experanto import Experiment`).

### Changes
- Added module docstring describing Experanto’s purpose and main exports
- Imported key classes:
  - `Experiment` (load and query multi-modal experiment data)
  - `ChunkDataset` (PyTorch Dataset for training)
  - `Interpolator` (base class for modality interpolators)
- Defined `__all__` to explicitly declare the public API
- Added `__version__ = "0.1.0"`

### Motivation
The empty `__init__.py` prevented users from accessing core classes via top-level imports. Defining `__all__` makes the package easier to use, improves discoverability, and clarifies the intended public API.

### Impact
- Users can now write cleaner imports:
  ```python
  from experanto import Experiment, ChunkDataset, Interpolator

### DIFF
--- a/experanto/__init__.py
+++ b/experanto/__init__.py
@@ -1,0 +1,20 @@
+"""
+Experanto: Interpolating recordings and stimuli in neuroscience experiments.
+
+Main exports:
+    - Experiment: Load and query multi-modal experiment data
+    - ChunkDataset: PyTorch Dataset for training
+    - Interpolator: Base class for modality interpolators
+"""
+
+from .experiment import Experiment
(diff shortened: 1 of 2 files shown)


pollytur:
this PR has changes which duplicate (conceptually) #126 
hence, closing it as there are changes outside of stated scope

Please do not reopen a PR with just adding the init.py file as this is a minor change

PR #161, opened 2026-04-15
fix: linear-interpolation

### LINKED ISSUE #160: start_time missing from times_lower/times_upper in linear interpolation
In `SequenceInterpolator.interpolate` and `PhaseShiftedSequenceInterpolator.interpolate`, `times_lower` and `times_upper` are computed without adding `start_time`

idx_lower is derived by `idx_lower = np.floor((valid_times - self.start_time) / self.time_delta).astype(int )`
but times_lower , is calculated as `times_lower = idx_lower * self.time_delta` 

for start_time != 0 it'll be innacurate in the numerator 
```python
lower_signal_ratio = ((times_upper - times_valid) / denom)[:, None] 
upper_signal_ratio = ((times_valid - times_lower) / denom)[:, None]
```

fix would be: 
```python
times_lower = self.start_time + idx_lower * self.time_delta 
times_upper = self.start_time + idx_upper * self.time_delta
```

Happy to open a PR if this is something that's to be addressed!

binary69:
Closes #160

Fixes the linear interpolation in `PhaseShiftedSequenceInterpolator.interpolate` and `SequenceInterpolator.interpolate`


binary69:
hey @pollytur @schewskone did the changes, happy to get a review :)

schewskone (approved):
LGTM, thanks for fixing this.

experanto/interpolators.py
```diff
@@ -447,16 +447,26 @@ def interpolate(
...
+            idx_upper = idx_upper[mask]
+            idx_lower = idx_lower[mask]
+            valid_times = valid_times[mask]
 
-            times_lower = idx_lower * self.time_delta
-            times_upper = idx_upper * self.time_delta
+            times_lower = (
+                self.start_time
+                + (idx_lower * self.time_delta)
+                + self._phase_shifts[np.newaxis, :]
```
pollytur: @binary69 why are there  self._phase_shifts[ here now while they were not there before

experanto/interpolators.py
```diff
@@ -447,16 +447,26 @@ def interpolate(
...
+            )
+            times_upper = (
+                self.start_time
+                + (idx_upper * self.time_delta)
+                + self._phase_shifts[np.newaxis, :]
```
pollytur: same here - @binary69 why are there  self._phase_shifts[ here now while they were not there before

experanto/interpolators.py
binary69: I just moved it there because the earlier version was just doing the same calculation in additional lines. Changed it since it'll make the code more readable

binary69: Thanks @pollytur I just made that change, from the surface level as it made it easier without time_dim and after looking through more carefully brought it back as it conceptually made sense. Sorry for the mess:(

pollytur (approved):

PR #164, opened 2026-07-05
Fix nearest_neighbor interpolation rounding (#163)

### LINKED ISSUE #163: keep_nans=False ignored in nearest_neighbor mode
Hi,

I noticed that SequenceInterpolator doesn't actually handle NaNs when using nearest_neighbor mode, even if keep_nans=False is set.

In interpolators.py, the linear mode has a check to fill NaNs using nanmean, but the nearest_neighbor block just returns the raw data from idx_lower:
if self.interpolation_mode == "nearest_neighbor":
    data = self._data[idx_lower]
    return (data, valid) if return_valid else data

because of this, interpolate() still returns NaNs when it's supposed to fill them. Is this a bug or am I missing something?

Also, idx_lower uses np.floor, so it’s always looking at the previous index instead of the actual "nearest" one. Just wanted to point that out.

sametSeckiiin:
Currently, `nearest_neighbor` interpolation uses `np.floor`, which ends up mapping timestamps to the previous index rather than the actual nearest neighbor. For example, querying `0.9s` (with a `1.0s` delta) returns the data from `0.0s` instead of `1.0s`.

This PR replaces `np.floor` with `np.round` so it correctly resolves to the closest time index. I also added `np.clip` to make sure rounding up doesn't cause an out-of-bounds error at the end of the array.

Fixes #163

Tested locally by querying intermediate timestamps to verify they pull from the nearest data point.


sametSeckiiin:
Hi @pollytur, whenever you have a moment, could you take a quick look at this PR? It fixes the nearest_neighbor interpolation logic discussed in #163. Thanks!

sametSeckiiin:
Hi @reneburghardt , whenever you have a moment, could you take a quick look at this PR? It fixes the nearest_neighbor interpolation logic discussed in https://github.com/sensorium-competition/experanto/issues/163. Thanks!

reneburghardt:
Hi @sametSeckiiin we are all not full-time working on Experanto and hence most of the time need a bit until we can answer and/or review PRs. 

Regarding this PR, I actually thought the original behavior was intended to avoid leakage from future data points (though we are arguably not fully consistent with this wrt linear interpolation anyways).

sametSeckiiin:
@reneburghardt No worries at all about the timing!

That makes total sense regarding the future data leakage. My thought process was just that people usually expect "nearest" to mean the mathematically closest point, while `floor` acts more like a "previous neighbor" step.

Since linear interpolation already looks ahead anyway, what do you think is best here? We could either stick with `np.round` for standard nearest behavior, or keep `np.floor` and maybe just clarify in the docs/naming that it avoids future leakage. 

Let me know which direction you prefer and I'll update the PR!

